PR #92, opened 2015-01-30
Logo beta tag

m4dz:


jsilvestre:
It looks good to me. @aenario I think you are the platform's TR?

aenario:
Beta is not translated by Polyglot, but is the same world for our two language so its fine.

aenario:
Oups, didnt notice, but the full build is 
cd client && brunch b && cd .. && cake build 

Otherwise your change dont appears when the server is started with node build/server.js as we do in production. I have completed the build.

m4dz:
> Oups, didnt notice, but the full build is `cd client && brunch b && cd .. && cake build`

Damned! Ok, thanks for the feedback. Maybe we can have a simpliest way to do this, or is there a reason to keep those build sub-commands separated?

jsilvestre:
It's because the build process is not normalized between apps. The most up-to-date apps have a single "cake build" command that does everything (Files has it).

m4dz:
OK, thanks :smile:!

PR #96, opened 2015-02-12
Fonts and icons

m4dz:
This PR:
- add cozy-fonts files and pre-setted CSS
- add happycloud favicon and apps icons


m4dz:
Oups, forgot the build, wait a minute :)

m4dz:
Ok for me. Let's review it, guys!

aenario:
Ok, after the ^ removal everything LGTM.
I like what you did with the various icons. Did you consider how to do the same in every apps (jade include/mixin ?)

m4dz:
Ok for the `^` :).

About icons, I have just added the FontAwesome pack in its 4.3.0 version. But this is not a real good choice for me because the whole fontawesome icons font-file will be downloaded by users, even if we just have 2 or 3 icons in use in the app.

So this choice is temporary, and I prefer to build a lighter subset version (using icomoon app) and embed the font-file base64-encoded directly in the `fonts.css` file. It'll be more efficient. I'll also write a guideline on how to use icons. I think classes are sufficients, but if you have some examples where a jade mixin is relevant, I'll make them with pleasure :)

m4dz:
Ok, I just understand what you say about _icons_… It's about favicons/appicons, right ?

aenario:
Yep, sorry i wasnt more explicit.

Regarding fontawesome, it will be downloaded only once, so I think it is better to include more icons (they can/will be used in all apps).

Regarding favicons/appicons, I think its ok to keep the proxy (happycloud) one  when installing as app with  romain.cozycloud.cc/#apps/calendar because when you open it, it will go to the full home and you can navigate amongs apps. I think we should actually set the app launch to be the home screen.

However, it would be nice to be able to install apps directly from romain.cozycloud.cc/apps/calendar with the app's icon and have it open directly to this app. For this to work, we will need to : 
A - duplicate the work you've done here of creating x versions of every app's icons
B - put the same code you've put in base.jade in every apps
I was trying to figure out a quick-win way to share B across apps, but it will only add complexity, better to explicitly do it manually.

For A, did you use a tool or did you do them all manually ? 

Anyway, this is all considerations for the future and dont change the fact that this PR is good to merge

m4dz:
Sounds good to me! As I [mentionned id in comments](https://github.com/cozy/cozy-proxy/pull/96/files#diff-bf0f5a369754c5445e5542daccbba9a7R9), I've used [RealFaviconGenerator](http://realfavicongenerator.net/) to build icons and manifest files. As I'll go through all apps to set fonts, I'll also set favicon in the same PRs :wink:.

aenario:
You are the best :trophy: 
Travis failed because of timeout, i restarted it. Good to merge once we have some green.

frankrousseau:
That looks interesting, but my the blue of the happy cloud seems not to matche the official one: http://cozy.io/assets/images/bighappycloud.png

m4dz:
> but the blue of the happy cloud doesn't seem to match the official one

I tried to recreate it as I haven't the official one in a large (1024*1024) version. I already have an SVG version, I just lack to correct gradient. Could you give it to me? Thanks :smile:!

frankrousseau:
I would be glad to have the SVG version. If you could provide it to me that would be nice. About the gradient you can use the top and bottom colors of the icon I mention above.

m4dz:
Ok, nice :) ! I wasn't about the orientation of the gradient, I'll do it this way.

frankrousseau:
:)

PR #101, opened 2015-03-06
Update README.md

cylwin:
Hello,

I add a way to hack the proxy using the cozy-vagrant. It was difficult for me to set up a dev env(3 hours) when I tried to hack some features of the proxy.

Bye

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,9 @@ Each modification requires a new build, here is how to run a build:
 
 ### To hack cozy-proxy using the cozy vagrant
 
-- Forward a new port from your vagrant (for example : `config.vm.network :forwarded_port, guest: 9555, host: 9555`)
+- Forward a new port from the virtual machine (for example: `config.vm.network :forwarded_port, guest: 9555, host: 9555` in file Vagrantfile)
 - Go in the shared folder `cd /vagrant` and `cd your-cozy-proxy-folder`
+- `rm -rf node_modules/bcrypt && npm install`
 - Launch cozy-proxy `PORT=9555 HOST="0.0.0.0" coffee server.coffee`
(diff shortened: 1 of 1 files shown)


frankrousseau:
@jsilvestre does it sound good to you?

jsilvestre:
That's actually the way to do it properly (I only have a few remarks). I think we should have by default port 9105 forwarded by the Vagrantfile, and see if we can do something to automatically seek for port 9105 on host machine (like we do in deploy) so it's just a "coffee server.coffee". I don't have time to test that so we'll see later.

Thank you @cylwin.

README.md
```diff
@@ -35,6 +35,13 @@ Each modification requires a new build, here is how to run a build:
 
     cake build
 
+### To hack cozy-proxy using the cozy vagrant
+
+- Forward a new port from your vagrant (for example : `config.vm.network :forwarded_port, guest: 9555, host: 9555`)
+- Go in the shared folder `cd /vagrant` and `cd your-cozy-proxy-folder`
```
jsilvestre: can you add that a "rm -rf node_modules/bcrypt && npm install" is needed too?

README.md
```diff
@@ -35,6 +35,13 @@ Each modification requires a new build, here is how to run a build:
 
     cake build
 
+### To hack cozy-proxy using the cozy vagrant
+
+- Forward a new port from your vagrant (for example : `config.vm.network :forwarded_port, guest: 9555, host: 9555`)
```
jsilvestre: * typo: "for example: " (no space before ":" in english)
* I suggest: Forward a new port from the virtual machine (for example: `config.vm.network :forwarded_port, guest: 9555, host: 9555` in file Vagrantfile)

PR #107, opened 2015-04-06
Improve styles

frankrousseau:
I performed the following changes:

* better responsive (smaller fonts and logo on the left for small screens)
* change font from Maven to Source Sans in most cases
* add a white background to the option field
* add a padding to the password field


aenario:
ping @m4dz  for review, 
The 3 size of maven-pro and 2 of signifika where removed from the "internal CDN", are they used in the other apps ?

m4dz:
It's ok for me. Fonts have been removed from previous commits, maybe due to an old upstream branch ;)

frankrousseau:
Thx to both of you!

PR #109, opened 2015-04-08
Improve English locale

jankeromnes:
This locale was already pretty good, but I couln't resist tweaking it some more!

My main problem was with "Send informations", which is incorrect (information is already plural) and misleading / scary: As a new user, I wondered "Where is my private information being sent to? Isn't the point of Cozy to not send information to the cloud?" Etc. I originally wanted to replace it with "Save", but "Finish registration" is closer to the French locale.

### DIFF
--- a/client/locales/en.coffee
+++ b/client/locales/en.coffee
@@ -3,98 +3,98 @@ module.exports =
     # login page
     "login title": "Cozy - Sign in"
     "login headline": "Welcome"
-    "login connection invitation": "Please type your password to sign in:"
+    "login connection invitation": "Please enter your password to sign in:"
     "login password placeholder" : "Your password"
     "login button" : "Sign in"
     "login forget password button" : "Did you forget your password?"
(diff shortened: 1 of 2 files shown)


client/locales/en.coffee
```diff
@@ -3,98 +3,98 @@ module.exports =
...
     # reset password page
     "reset title": "Cozy - Reset password"
-    "reset headline": "Tell me your new password"
-    "reset instructions": "After a password reset, you must reconfigure your confidential data like your mailbox or your bank account."
+    "reset headline": "Enter your new password"
+    "reset instructions": "After a password reset, you will have to reconfigure confidential data like your mailbox or bank accounts."
     "reset password placeholder": "Your new password"
     "reset button": "Change my password"
-    "reset success message": "Password successfully reset."
+    "reset success message": "Password reset successful."
```
frankrousseau: Are you sure about that one?

frankrousseau:
Great! Thank you @jankeromnes

PR #115, opened 2015-04-16
Try to retrieve localization several times

poupotte:


server/controllers/authentication.coffee
```diff
@@ -68,6 +68,22 @@ module.exports.register = (req, res, next) ->
...
+        if counter is 5
+            cb "Cannot recover polyglot"
+        else
+            polyglot = localization.getPolyglot()
+            if polyglot?.t?
+                cb null, polyglot
+            else
+                setTimeout () ->
+                    counter += 1
+                    recoverPolyglot cb
```
aenario: recover vs retrieve

poupotte: Done. Thanks @aenario

PR #122, opened 2015-05-13
Update fr.coffee

goofy-bz:
trying to correct a plain text html link in an interface message.

notre forum lien
arf_source


frankrousseau:
Thank you for this fix. But it will be soon fixed (we'll publish a new version of the Proxy and fix is included). See development branch for that changes.

PR #132, opened 2015-07-02
Fix build

m4dz:


m4dz:
Replaces #131, this one should be the good one (I hope)…

PR #136, opened 2015-07-06
Feedbacks and code cleaning

m4dz:
This PR contains enhancements for user feedback as well as code cleaning for a better and easier maintenance in code logics.


frankrousseau:
I merged this request but there are still problems with the success state. I asked you to change the button with green and a login success message. Instead I had a check icon and a "login undefined" message below. I prefer my initial requirement which is based on hundreds of login into my Cozy. It's better that we stick on it.

I tried to fix it myself but I failed at understanding the new architecture. So we'll have to fix it together.

PR #137, opened 2015-07-06
Leave Google progress patch

m4dz:
This PR contains a feedback enhancement for leave google, that block user in the setup screen with a progress bar for its imports before letting him go the the welcome screen and the home app.

This is to prevent users that go to home before all contents are fully imported from Google, and then open a realtimed app (like contacts) which creates issues in browsers (i.e. have a mass of contacts coming in realtime crashes the contact app).

This PR starts from #136, so merge this last before :smile:.

### DIFF
--- a/client/app/locales/en.coffee
+++ b/client/app/locales/en.coffee
@@ -50,6 +50,7 @@ module.exports =
     'login enter your password':    "Please enter your password to access to your Cozy"
     'login password':               "password"
     'login submit':                 "sign in"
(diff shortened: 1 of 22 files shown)


frankrousseau:
Thank you for your PR.

The error feedback looks to work. 

But the import doesn't seem to work (my import is blocked). Plus there is only one progress bar for all while one progress bar for each data type like in the Leave Google bar would make the waiting funnier. We'll work on it tomorrow.

PR #139, opened 2015-07-07
sign in feedback improvements

m4dz:
This PR improve the way sign in display its feedback to the user.


frankrousseau:
Thank you, it works good now!

PR #144, opened 2015-07-10
Fix progress bar style

m4dz:


frankrousseau:
Hi @m4dz,

Thank you for this fix. I think we should come back to the old process where we explain what we do in the Pull Request.

Can you add details in your PR explaining what are `autoprefixer` and `postcss`?

PR #163, opened 2015-08-20
Small fixes

frankrousseau:
* Select password when login fails.
* Enable sign in field when Firefox auto-fill the password field.
* Remove emails from import (it's not working)


frankrousseau:
It includes @nono's PR

PR #166, opened 2015-08-26
[NEED DISCUSSION] Specify a max-age for static assets 

aenario:
There is a tradeof here : 

By default, the max-age is 0 and the proxy assets can never be totally cached. 

In emails, we create an iframe with a link to the /fonts/fonts.css every time a message is displayed, this cause the browser to make 3 requests (fonts.css & 2*.woff) and recompute fonts even if it receive 304 responses.
  
However, the usage of high max ages is most probably the cause of our cache problems (forcing users to ctrl + F5 after an update).

For now, i think i will stop cozy-emails from using the proxy "cdn" as the use case of creating iframes is very specific.

Let's make this the discussion about the cache problem and how best to solve it not only for proxy but for all our apps.
Pinging @m4dz @clochix @frankrousseau @jsilvestre

### DIFF
--- a/server/config.coffee
+++ b/server/config.coffee
@@ -33,7 +33,9 @@ config =
             americano.errorHandler
                 dumpExceptions: true
                 showStack: true
-            americano.static path.join __dirname, '/../client/public'
+            americano.static path.join __dirname, '/../client/public',
+                maxAge: 5 * 60 * 1000
+
             selectiveBodyParser
             usetracker
             authSteps[0]


frankrousseau:
From what I read there in your message, there are only good reasons for that. Can you tell more about the cons of this feature?

aenario:
My bad, i was not clear : 

Option 1 - Close this PR, keep max-age at default 0
Brower will always ask the server for all proxy assets, the server will reply with 304 and no body most of the time. 
This can cause flash of fonts if the connection is slow.
This probably fix the bug after update for the proxy only. 
Emails keeps flashing at each message opening (but this can be fixed in emails)

Option 2 - Merge this PR, increase max-age
Browser will not ask the server for new version of proxy asset if the browser downloaded it less than 10 minutes ago
This probably will bring back the bugs after update of the proxy (unless we do some brunch-hash magic)
Email can keep using the proxy cdn fonts (not very important)


So, my recommendation would be to not merge this PR, fix the emails bug in emails and add an issue in all applications to set maxage to 0.

clochix:
@aenario What is the status of this PR ? Are you waiting for something ?

PR #171, opened 2015-09-08
Display error if proxy doesn't access to user

poupotte:


server/controllers/authentication.coffee
```diff
@@ -9,9 +9,15 @@ localization = require '../lib/localization_manager'
...
 
-module.exports.registerIndex = (req, res) ->
+module.exports.registerIndex = (req, res, next) ->
     User.first (err, user) ->
-        if user?
+        if err?
+            error = new Error err.code
+            error.status = 500
+            error.template =
+                name: 'error'
```
frankrousseau: can you make the error message more explicit?

poupotte: For user, proxy displays : 
error_proxy
Do you want I create a specific error page when proxy can't retrieve user ?

frankrousseau:
Ok let's use it that way.

PR #176, opened 2015-09-17
Migrate lang files to Transifex

clochix:


frankrousseau:
Thank you @lemelon and @clochix

Issue #174, opened 2015-09-15
Proxy submit register request multiple times


m4dz:
When registering, the registration endpoint is called multiple times, resulting of 409 errors.

**Reproduce**
1. register a new user
2. pass the leave google form (registration is sent again > 409)
3. pass email configuration form (registration is sent again > 409)

**Expected result**
The form must not be sent when passing form in next steps.

clochix:
@m4dz this should have been fixed. Could you give the patch a try and close this ticket, please?
Thanks for reporting. Thanks @poupotte for fixing ☺

m4dz:
Ok for me, thanks again @poupotte :smiley:!

PR #191, opened 2015-10-07
Don't submit register request multiple times, refs #174

poupotte:


frankrousseau:
Thank you  @poupotte for your PR. 

There is a problem here. If an error occurs, the user cannot signup anymore.

PR #193, opened 2015-10-15
Use http code 204, better for a deleted content

nono:
Hi,

I think it should be better to send a 204 No Content when a device is deleted. It matches more elegantly the HTTP semantics and it avoids to send `OK`, which is a not valid JSON payload.

204 No Content

### DIFF
--- a/server/controllers/devices.coffee
+++ b/server/controllers/devices.coffee
@@ -255,7 +255,7 @@ module.exports.remove = (req, res, next) ->
                     if err?
                         next err
                     else
-                        res.send 200
+                        res.send 204
 
     initAuth req, (user) ->
         # Check if request is authenticated


frankrousseau:
:)

frankrousseau:
Let's make the PR!

nono:
@frankrousseau but, it's the PR!

frankrousseau:
Oups, sorry I thought it was an issue with commit reference :)

aenario:
We have an issue with our process : this PR broke the tests but because the test are run against the built version, travis said its OK until we actually try to build & publish it.

nono:
@aenario we can test the coffee, or add a build step to travis. Do you know why we test the JS files?

PR #200, opened 2015-11-04
Environment awareness

m4dz:
This PR introduces an environment aware onboarding workflow:
- transmit to client app the environment variables in a dedicated module
- skip the import from google step if the dedicated app isn't available on the cozy
- remove email setup step


server/controllers/authentication.coffee
```diff
@@ -9,23 +10,52 @@ localization = require '../lib/localization_manager'
 passwordKeys = require '../lib/password_keys'
 
 
+getEnv = (callback) ->
+    async.parallel [
+        (callback) ->
+            User.first (err, user) ->
```
aenario: I'd make all this into a User.getusername() function.

server/controllers/authentication.coffee
```diff
@@ -9,23 +10,52 @@ localization = require '../lib/localization_manager'
...
+                    username: user.public_name
+                else
+                    username: helpers.hideEmail user.email
+                        .split ' '
+                        .map (word) -> word[0].toUpperCase() + word.slice(1)
+                        .join ' '
+                callback null, env
+
+        , (callback) ->
+            env = apps: (key for key of require('../lib/router').getRoutes())
```
aenario: use Object.keys

server/controllers/authentication.coffee
```diff
@@ -9,23 +10,52 @@ localization = require '../lib/localization_manager'
...
+            callback null, env
+    ],
+    (err, results) ->
```
aenario: We don't really need the async.parallel function here, you might as well remove it, we can reintroduce it if we add more env.

server/controllers/authentication.coffee
m4dz: Hmmm, I'm not sure to understand what you mean :confused:

server/controllers/authentication.coffee
m4dz: Of course, damned!

server/controllers/authentication.coffee
m4dz: True. I just wanted to keep things separated to be more flexible, but I agree, I'll rewrite it.

PR #207, opened 2015-11-12
Make Google import more consistent

m4dz:


client/app/views/register/setup.coffee
```diff
@@ -83,28 +100,39 @@ module.exports = class RegisterSetupView extends Mn.ItemView
...
-            args = [getProgress, timer.toProperty()]
-            if 'contacts' in imports
-                args.push fromSocket('contacts').toProperty()
-            if 'calendars' in imports
-                args.push fromSocket('calendars').toProperty()
+            streams = imports.map (datasource) ->
+                fromSocket(datasource).toProperty()
 
+            args = [getProgress, timer()].concat streams
             @progress = Bacon.combineWith.apply Bacon, args
```
aenario: Bacon.combineWith getProgress, timer(), streams...

aenario:
There also appears to be some conflicts with locales following commit 9a23647, could you take a look ?

client/app/views/register/setup.coffee
m4dz: Oooohhhh :heart_eyes:

m4dz:
@aenario it's ok for me :+1:

PR #208, opened 2015-11-13
Improve timezone selector

m4dz:


client/app/views/templates/view_register_preset.jade
```diff
@@ -24,11 +24,19 @@ p#preset-email-tip.tips
...
+        - var skipstep  = false
+        - var lastcont = null
+        each timezone, index in timezones
+            unless skipstep
+                - var cont = timezone.split('/')[0]
+                if cont != lastcont
+                    - lastcont = cont
+                    if index
+                        option(disabled=true)= '---'
+                if cont.toLowerCase() == 'africa'
```
nono: @m4dz I'm a bit lost. What is the special case with africa?

m4dz: Its starts by a "A" :smile:
More seriously, we first presents 4 Europe's timezones, then 2 America's ones, and finally display all available timezones in alphabetical order. This trick is just to separate the 2 first blocks from the rest of the list.

nono: Okay...

PR #209, opened 2015-11-16
show static app

lemelon:
This PR proposes to show the static apps and to resolve some exceptions linked to the behavior of an app without a server


server/controllers/applications.coffee
```diff
@@ -1,20 +1,36 @@
...
             error.status = err.code
             error.template =
                 name: if err.code is 404 then 'not_found' else 'error_app'
             next error
+        else if result.port?
+            getProxy().web req, res, target: "http://localhost:#{result.port}"
         else
-            getProxy().web req, res, target: "http://localhost:#{port}"
+            # if public repository exists, then redirect to public app
+            if fs.existsSync result.path + '/public/index.html'
```
aenario: use require('path').join instead of string manipulation

server/controllers/applications.coffee
```diff
@@ -1,20 +1,36 @@
...
+                req.url = "/public/#{appName}"
+                res.redirect "#{req.url}/"
+            else
+                # showing private static app
+                appManager.startStaticApp appName, req.url, result.path, (
```
aenario: It would be better to use one of the many static assets handler middleware, or even res.sendFile. 
It avoid loading the whole index.html into memory, better manage request aborting and for the middleware case, can handle caching too.

server/lib/app_manager.coffee
```diff
@@ -59,6 +63,14 @@ class AppManager
                     state: data.app.state
                 callback null, data.app.port
 
+    # readfile to start a static app
+    startStaticApp: (appName, url, path, callback) ->
+        logger.info "Starting static app #{appName}"
+        url += 'index.html' if url is '/' or url is '/public/'
+        fs.readFile path + url, 'utf8', (err, content) ->
```
aenario: use path.join too, you also need to make sure one cant put  ".." in the URL to browse the whole filesystem.

lemelon: I think that I've managed to block the root path (if a user has the bad idea to link his app by doing href="../../../blog.html" for example). However i tested this situation and the answer of the proxy server is a 404 not found which sounds like logical for me. So i can't really see the use of this yet. Maybe there's something i did'nt consider. If anyone can help me with this particularity that would be great.

frankrousseau: Please with coffeescript don't use `+` to build your strings. It's better if you use `"#{var1}#{var2}"`.

aenario:
Ok, one more round of changes : 

- Please rename getUrlFor into getPathFor (url is in a network request, this is a path on the FS)

- Please try also with https://github.com/aenario/cozy-front-nopublic : http://my.cozycloud.cc/public/front-nopublic/ should get the `<h1>Hello world</h1>` page. (if there is no public folder, we consider its a single application serving both the public and internal role.)

- It's good to test manually, it's better to have travis testing for you Please add mocha tests with both tests repo.

Thanks !

frankrousseau:
Do not forget to add tests.

PR #211, opened 2015-11-20
Fix domain in password reset link

clochix:


server/controllers/authentication.coffee
```diff
@@ -103,7 +103,9 @@ module.exports.forgotPassword = (req, res, next) ->
             Instance.first (err, instance) ->
                 return next err if err
 
-                instance = domain: 'domain.not.set'
+                if not instance?
+                    instance = domain: 'domain.not.set'
```
nono: @clochix you can shorten it to:

```coffee
instance ?= domain: 'domain.not.set'
```

Issue #215, opened 2015-12-07
Missing url encoding of "?" when redirected to /login


khertan:
Missing url encoding of "?" when redirected to /login
This prevent registering a remotestorage from litewrite.net for example (you need to logout before).

clochix:
Thanks for reporting and fixing @khertan,

Your fix will be in the next release of the proxy, probably next week. Have you been able to test it on your server, or do you want some help to install it?

khertan:
Thanks, it's already installed at home :) I tested it before committing ;)

poupotte:
Thanks @khertan

PR #216, opened 2015-12-07
Correct urlencode ? for redirection from login when user isn t auth. …

khertan:
Fix issue #215

When user is not logged, the redirection to /login for authorizing remotestorage connection redirect to /login?next=/apps/remotestorage/oauth/?redirect_uri=http://litewrite.net/&scope=documents:rw&client_id=http://litewrite.net&response_type=token

the second ? should be urlencoded else the parameter of next= aren t present after login on the next redirection.

### DIFF
--- a/server/middlewares/authentication.coffee
+++ b/server/middlewares/authentication.coffee
@@ -33,6 +33,6 @@ module.exports.isAuthenticated = (req, res, next) ->
         next()
     else
         url = "/login"
-        url += "?next=#{req.url}" unless req.url is '/'
+        url += "?next=#{req.url.replace "?", "%3F"}" unless req.url is '/'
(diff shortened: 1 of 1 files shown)


frankrousseau:
Hello @khertan,

  I think you should use the `encodeURIComponent` javascript function to do that instead of `replace`.

Ex:

```coffeescript
encodeURIComponent 'http://mysite.com/?var=value'
```
returns:

```coffeescript
'http%3A%2F%2Fgoogle.com%2F%3Ftata%3Da'
```

```

khertan:
frankrousseau, that s what i think too the first time, but then you also encode things that should not be encoded because it s not decoded after. Or maybe we should decode them after in client

frankrousseau:
Great, thank you!

PR #222, opened 2015-12-15
resolve test for uriencodecomponent

lemelon:
In the previous PR, test had failed on uriencodecomponent param. This PR resolves this bug


nono:
Nice!

PR #226, opened 2016-01-04
add access for static app

lemelon:
This PR enables to initialize or update static apps and non-static apps. It's important for me to do this because in this case the static apps will be able to access the data related to the permissions.


server/models/device.coffee
```diff
@@ -20,25 +20,19 @@ if process.env.NODE_ENV is "production" or process.env.NODE_ENV is "test"
...
-                    if err?
-                        logger.error err
-                        callback err
-                    else
-                        for access in accesses
-                            # Check if access correspond to a device
-                            if access.key in devices
-                                cache[access.value.login] = access.value.token
+                for access in accesses
+                    if access.key in devices
```
frankrousseau: I don't see where devices is declared. Are you sure you want to base your condition on it?

lemelon: Var devices is declared in l31 but i was thinking that it could better to name it apps instead of devices because static apps are not necessarily devices. What do you think ?

frankrousseau: Oups sorry. Yes renaming it is better.

lemelon: Yes actually i think that i did'nt need to map it so you were right. If it's ok for you and for @poupotte we can merge it.

PR #229, opened 2016-01-06
Add a postinstall script for node4 compat

nono:
Bcrypt@0.8.1 is compatible with node 0.x.y
Bcrypt@0.8.5 is compatible with node 0.10.40+

We want cozy-proxy to works on:

- node 0.10.25 (the version installed by default on ubuntu)
- node 0.10.29 (the version installed by default on debian)
- the current stable release of node 0.10, 0.12, 4 and 5

So, we have to use the postinstall feature of npm to check the version
of node and install a version of bcrypt compatible with it.

### DIFF
--- a/postinstall.js
+++ b/postinstall.js
@@ -1,0 +1,10 @@
+#!/usr/bin/env node
+
+var spawn = require('child_process').spawn;
+var major = process.versions.node.split('.')[0];
+var bcrypt = 'bcrypt@0.8.5';
+
+if (major == '0') {
+  bcrypt = 'bcrypt@0.8.1';
+}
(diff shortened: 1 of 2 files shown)


postinstall.js
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
+
+var spawn = require('child_process').spawn;
+var major = process.versions.node.split('.')[0];
+var bcrypt = 'bcrypt@0.8.5';
+
+if (major == '0') {
```
frankrousseau: With JS it's more common to use triple `=` like `===`.

nono: You are right. Too much coffeescript and I forgot the JS best practices!

frankrousseau:
Thank you @nono

PR #235, opened 2016-01-26
Avoid one source of sockets in CLOSE_WAIT state

nono:


nono:
Travis still runs the build with js files but doesn't compile them itself...

frankrousseau:
@nono Your PR breaks the tests.

server/controllers/applications.coffee
```diff
@@ -12,41 +12,36 @@ getPathForStaticApp = (appName, path, root, callback) ->
...
 
+module.exports.app = (req, res, next) ->
+    req.url = req.url.substring "/apps/#{appName}".length
+    errTemplate = (err) ->
+        name: if err.code is 404 then 'not_found' else 'error_app'
+    forwardRequest req, res, errTemplate, next
+
 module.exports.publicApp = (req, res, next) ->
-    appName = req.params.name
     req.url = req.url.substring "/public/#{appName}".length
```
aenario: you remove appName but it's still used here

server/controllers/applications.coffee
```diff
@@ -12,41 +12,36 @@ getPathForStaticApp = (appName, path, root, callback) ->
...
-                name: 'error_public'
-            next error
-        else if result.type is 'static'
-            # showing public static app
-            getPathForStaticApp appName, req.url, result.path, (url) ->
-                send(req, url).pipe res
-        else
-            getProxy().web req, res, target: "http://localhost:#{result.port}"
+    errTemplate = (err) ->
+        name: 'error_public'
```
aenario: are you not supposed to call forwardRequest here ?

server/controllers/applications.coffee
nono: @aenario thanks for doing my debug

server/controllers/devices.coffee
```diff
@@ -61,7 +61,7 @@ deviceExists = (login, cb) ->
         else if body.length is 0
             cb null, false
         else
-            cb null, body[0]
+            cb null, body[0]?.value
```
nono: @aenario I have no idea what have changed that causes this. Does it ring a bell for you?

aenario: No idea, I checked the DS & Cradle too and couldn't find anything. 
Was this caught by travis ?

nono: Yes, for example here: https://travis-ci.org/cozy/cozy-proxy/jobs/104924355#L753

aenario: I think it is one if these case where things just happened to work : 

body[0] would be `{id: "xxx", key:"xxx", value:{actual doc}}`, we return this in callback as device.
It is truthy, so the `if device` in `checkLogin` will pass, then we do `device.id` in removeDevice, which pass too.

So this is a good change, but i dont think it was this one which fixed the error in travis :)

nono: It is the source of the bug on travis. `checkLogin` is used for getting the actual doc when updating. Without it, the updated doc lost its `login` field. And when we try to delete the device, it is no longer found, as he has no login.

aenario: Well then, One more "I have no idea how this ever worked"

PR #238, opened 2016-02-04
Improve the fix for sockets in CLOSE_WAIT

nono:


aenario:
Good to merge when travis is happy (just restarted, github fetch failure)

Issue #199, opened 2015-10-31
Add warning when JS or cookies are disabled


clochix:
(feedback from a user)

Some of our users take really care of their online privacy, and block cookies and JS by default. We should try to detect this case and display a warning message.

poupotte:
This issue is fixed in new version : 2.2.0

Issue #154, opened 2015-07-31
Login page do not redirect to home (when already logged in)


PCYoshi:

clochix:
@PCYoshi means that when user is already connected and navigate to the login page (for example, from it's browser bookmarks), we should redirect her to the home page.

poupotte:
This issue is fixed in new version : 2.2.0

PR #239, opened 2016-02-05
Fix for #154 and #199

nono:


server/views/index.jade
```diff
@@ -7,6 +7,13 @@ block title
 
 block content
     main(role='application')
+      noscript
+        section.popup
+          header
+            a(title='Cozy Website', href='https://cozy.io')
+          .container
+            h1 Sorry
+            p.help Cozy needs JavaScript and cookies to work but it looks like you have disabled one of them.
```
frankrousseau: Shouldn't it be localized?

nono: The title and description above in this page are not localized. So, I've followed their example. But it would probably be better to localize them.

clochix:
Please don't forget to push to Transifex every time you update a source locale file ;-)

PR #244, opened 2016-02-17
Make node_modules lighter in production

nono:
- Move npm-run-all to devDependencies
- Replace americano-cozy by cozydb


nono:
Travis build is red for node 4&5, but it's not my fault. It failed on `brunch build --production` but I've only modified the server part, not client. @m4dz will check if we can upgrade brunch and its modules.

m4dz:
I''m on it, I try to fix some issues with Brunch 2…

PR #245, opened 2016-02-18
Fix #244

m4dz:
This one is continuity of #244 that try to fix build errors on node v4 / v5


frankrousseau:
Thank to both of you!

PR #246, opened 2016-02-20
[POC] Webpack front build

m4dz:
This PR introduces Webpack for frontend build. It supports in this current configuration:
- build app JS and CSS files
- fast compilation
- all frontend deps from NPM
- static assets support
- code spliting for onboarding section
- optimization in `production` build mode
- watch mode for simplified development

Pros:
- efficient compilation
- clever piping system
- code spliting that produces chunks which are loaded dynamically when needed (in this case, _onboarding_ code is split in a dedicated chunk which reduce the main codebase of ~25%)
- quick rebuild on watcher (still impressive)

Cons:
- webpack config file is verbose, even for a _basic_ context
- it needs to understand webpack _loaders_ and _plugins_ concepts to really take advantages of the webpack pipeline system

My thoughs: even with their efforts, Brunch@2.x is still unstable with NPM deps, and sometimes make corrupted builds on watch mode. I didn't encounter this kind of issues with webpack (very fast and stable) and I think the final codebase is more robust (all deps declared, splitted code, etc). The config stay verbose, but can be easily deployed across our apps in a _standard_ base config and can be customized after.

This PR is fully functionnal but serves as a POC for next ideations :smiley:.


client/app/routes/index.coffee
```diff
@@ -40,6 +36,9 @@ module.exports = class Router extends Backbone.Router
     the submitted form.
     ###
     auth: (path, options) ->
+        AuthView  = require '../views/auth'
+        AuthModel = require '../states/auth'
```
nono: Why did you put the `require`s here and id not kept them on top of the file?

m4dz: I split the code concerning the Registration (onboarding) part below, and mode a lazy-require. So I put the _Auth_ requires here to be consistent, but it's trully only cosmetic :smile:

package.json
```diff
@@ -18,31 +18,29 @@
...
     "graceful-fs": "3.0.8",
     "http-proxy": "1.12.0",
-    "jade": "1.11.0",
     "locale": "0.0.14",
     "lockedpath": "0.1.1",
     "node-polyglot": "0.3.0",
     "passport": "0.2.0",
     "passport-local": "1.0.0",
     "printit": "0.1.15",
+    "pug-runtime": "2.0.0",
```
nono: We are using `jade 1` below and `pug-runtime 2` here. It's not the same major version. Is it OK to do so?

build/server/middlewares/authentication.js
```diff
@@ -1,70 +0,0 @@
-// Generated by CoffeeScript 1.10.0
```
nono: Why did we lost the `build/server` directory?

server/config.coffee
```diff
@@ -33,7 +33,8 @@ config =
             americano.errorHandler
                 dumpExceptions: true
                 showStack: true
-            americano.static path.join __dirname, '/../client/public'
+            americano.static path.join __dirname, '../static'
```
nono: Why `client/public` was moved to `static`? I think it was logic to have client-side assets in the `client` folder.

server/views/index.jade
```diff
@@ -16,10 +16,7 @@ block content
...
 block js
-    script(src="/scripts/vendor.js")
     if env
         script.
-            require.register('env', function(exports, require, module) {
-              module.exports = !{JSON.stringify(env)}
-            })
-    script(src="/scripts/app.js", onload="require('initialize');")
+            ENV = !{JSON.stringify(env)};
+    script(src="/app.js" defer=true)
```
nono: Webpack does not offer something for loading assets with a unique name? It's convenient to set far future expires headers.

package.json
m4dz: Seems that's the pug-runtime (formerly [jade-runtime](https://www.npmjs.com/package/jade-runtime), this renamming dance is madness) is still full compatible with jade@1. I didn't seen any regression during my tests.

server/views/index.jade
nono: Probably related to https://webpack.github.io/docs/long-term-caching.html

build/server/middlewares/authentication.js
m4dz: I though that with only keep build directory on `master` branch to keep all dev branches clean and prevent commit mess with useless diff concerning build folder. Nope?

client/webpack.config.js
```diff
@@ -0,0 +1,85 @@
...
+    },
+    resolve: {
+        extensions: ['', '.js', '.coffee', '.jade', '.json']
+    },
+    debug: !production,
+    devtool: production ? false : 'eval',
+    module: {
+        loaders: [
+            {
+                test: /\.coffee/,
```
nono: I should be better to use `/\.coffee$/`, no?

build/server/middlewares/authentication.js
nono: It's OK for me.

@aenario what do you think of it?

server/config.coffee
m4dz: This one is a little bit tricky : we can put all frontend assets in the client folder, but they need to be referenced in modules to be built by webpack. It's ok with img, stylesheets, etc. But we also need assets related to the app (here, the favicons and manifests) which are only used in the index page served by the server-part.

So I though it'll be better to:
1. have a `static` folder at root whoch contins all statics stuff related to the whole app (and be copied as-it on build)
2. have another folder in client part containing all stuff linked from modules

I think it's a good and flexible approch, but I'm totally aware to any improvements :smile:

server/views/index.jade
m4dz: Webpack offers a way to append a hash to generated files, but I ask myself if using a cache-manifest (using https://github.com/lettertwo/appcache-webpack-plugin) wo'nt be a better way to do this…

client/webpack.config.js
m4dz: True :+1:

server/views/index.jade
nono: https://developer.mozilla.org/en-US/docs/Web/HTML/Using_the_application_cache advises to avoid appcache, as it is deprecated and can be a pain to use/debug. Service Worker is recommended now.

frankrousseau:
@m4dz Could you remove the build files. It makes the review hard to do.

m4dz:
@frankrousseau I've suppressed the build removing from the PR. I'll open another PR for the clean (build and locales) as long as we don't push build and locales to other branches than master anymore.

server/config.coffee
frankrousseau: Please don't change too much things at a time. We have a clear convention for client and server folders in all our folders. It's going to be a total mess, there are so already so many differences between projects.

frankrousseau:
Thank you for removing the build. About the folder name change, I prefer we stick to our current convention. Including too many changes at the same time will create more chaos. Introducing webpack already brings enough changes.

m4dz:
About static folder convention, Webpack will copy assets only if are referenced by a module (there's a "copy plugon" but it wll introduce mess in our files management if we mix in the same folder assets used by modules and those whose aren't). So, I think we'll need to keep those assets separately. We can continue to produce a `client/public` folder for hosting assets (instead of my `static` choice) but where can we keep assets not refercned (like _favicons_ or _fonts_)? I suggest:

1. use a `public` folder at root that contain them (same as now, just rename `static` to `public`)
2. use a `client/vendor` folder from which all `assets` folder(s) will be copied directly to the `build/public` (a-la-brunch) (maybe using the copy plugin).

I think the 2. is a little bit more complicated but is the more reliable from our current use in *brunch*ed projects. @frankrousseau, any pref?

frankrousseau:
I want to keep everything related to browser in the client folder. Subfolders can organized the way you want, but please keep the client and server separation. 

Currently, I don't have much time to review the full thing. But from what I understand the second solution looks to follow our current convention. If that's the case, ok for it.

aenario:
#247 was merged, could you rebase this one ?

m4dz:
Done!

client/webpack.config.js
```diff
@@ -0,0 +1,99 @@
...
+        })
+    ]);
+}
+
+module.exports = {
+    entry: './app/initialize',
+    output: {
+        path: path.join(production? '../build/client' : '', 'public'),
+        filename: production? 'app.[hash].js' : 'app.js',
+        chunkFilename: production? 'register.[hash].js' : 'register.js'
```
nono: @m4dz what is this `register.js` file?

nono:
Looks good to me.

Before: 367K (87K for app.js + 280K for vendor.js)
After: 244K for app.[hash].js

Nice size reduction. Thanks @m4dz!

client/webpack.config.js
nono: No sourcemap for production?

m4dz: I use the code splitting feature in Webpack to make a chunk for the onboarding (register) process. It decrease the app.js file by 96K, and is obviously only use one time at first access. Webpack lazy load the `register` chunk when accessing the `/register` url, so it's totally transparent for users.

m4dz: We have disabled sourcemaps in prod with brunch, not sure they're really useful. Do you think we should enabled them?

aenario: We should enable them in a separate file if possible.

nono: Webpack has some nice features :smile: 

For sourcemaps, I don't know. I'm not the one who does most front-end code.

m4dz:
> Nice size reduction

I also ask myslef to enable compression using gzip in proxy server part, but as long as proxy serves all apps responses, it means enable it for all apps at one time. Should be ok, but I'll try it later in another PR to check if it's really good :smile:.

client/webpack.config.js
```diff
@@ -0,0 +1,99 @@
+var path = require('path');
+
+var webpack           = require('webpack');
+
+var ExtractTextPlugin = require('extract-text-webpack-plugin');
+var CopyPlugin        = require('copy-webpack-plugin');
+var AssetsPlugin      = require('assets-webpack-plugin');
+var BrowserSyncPlugin = require('browser-sync-webpack-plugin');
+
+var production = process.env.NODE_ENV === 'production';
```
aenario: The situation with NODE_ENV is already a bit complicated with vm-dev in production or development mode and actual production. Could we avoid adding build-time env to the mix ?

Can we use a --flag or another env name ?

frankrousseau: Sorry I didn't see your comment aenario.

server/views/index.jade
```diff
@@ -16,10 +16,7 @@ block content
...
 block js
-    script(src="/scripts/vendor.js")
     if env
         script.
-            require.register('env', function(exports, require, module) {
-              module.exports = !{JSON.stringify(env)}
-            })
-    script(src="/scripts/app.js", onload="require('initialize');")
+            ENV = !{JSON.stringify(env)};
+    script(src="/#{assets? assets.js : 'app.js'}" defer=true)
```
aenario: I would move this logic to server / initialize.

server/views/layouts/_base.jade
```diff
@@ -26,9 +26,10 @@ html(lang=getLocale())
         meta(name="msapplication-TileImage", content="/mstile-144x144.png")
         meta(name="theme-color", content="#20a8f1")
 
-        link(rel='stylesheet', href='/fonts/fonts.css')
-        link(rel='stylesheet', href='/styles/app.css')
+        link(rel="stylesheet", href="/fonts/fonts.css")
+        link(rel="stylesheet", href="/#{assets? assets.css : 'app.css'}")
```
aenario: idem than for JS

client/webpack.config.js
m4dz: As far as I know, Webpack doesn't offer a flag to enable a specific mode (only a flag to explicitely activate optimizations) and rely on `NODE_ENV=production` to atomatically turn optimnizations on.

This one is only use at build time (see npm script `build:client`), but we can imagine use another ENV var to enable those parts if we want. What do you suggest to use?

server/views/index.jade
m4dz: It's already the case in `initialiaze` when we try to load the `webpack-assets.json` file. Do you prefer we define a fallback explicitely instead of in templates?

client/webpack.config.js
aenario: if the NODE_ENV also has an impact on webpack, then leave it as is.

m4dz: Ok, nevermind, we can switch the env vars to `OPTIMIZE=true` and do it manually as we already done. Let's do it!

server/views/index.jade
aenario: exactly

PR #247, opened 2016-02-22
[chore] clean useless build files

m4dz:


m4dz:
@nono is this one ok for you? So We can merge it and I rebase the #246 on it to simplify the review.

nono:
We should ask our release manager (ping @aenario). If we do so, the master and development branches will be different. In particular, the `.gitignore` file will ignore the build on development, but not in master. I'm not sure about that change.

aenario:
That's what we have in emails and while it is a bit of hassle for merging, it avoids meaningless conflicts in development. Let's discuss in TC.

nono:
OK, let's go for that!

PR #250, opened 2016-02-25
Continue #246

m4dz:
This PR continue #246 by:
- fix README doc
- use env `OPTIMIZE=true` instead of `NODE_ENV=production` to switch to production build
- enable sourcemaps everywhere
- explicitely define assets in dev mode
- improve webpack config for readability and doc

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,9 @@ We assume here that the Cozy platform is correctly [installed](https://raw.githu
 
 Type this command to install the proxy module:
 
-    cozy-monitor install proxy
+```sh
+$ cozy-monitor install proxy
+```
 
 ## Contribution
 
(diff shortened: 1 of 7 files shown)


frankrousseau:
@aenario I let you review and merge it. Sorry for the previous merge.

frankrousseau:
README is still not updated. I still can't build my client because I don't know how to do it.

frankrousseau:
I mean that README : https://github.com/cozy/cozy-proxy/blob/master/client/README.md

m4dz:
Damned! I forgot the client README :disappointed:, sorry…

m4dz:
@frankrousseau is it better?

aenario:
What's your logic for separation between vendor/assets and app/assets ?
We might as well drop the vendor folder entirely.

frankrousseau:
@m4dz That's much better. Could you make the file < 80 chars wide? At first it looks complicated but with JS, you can add carriage returns easily I think.

m4dz:
> What's your logic for separation between vendor/assets and app/assets?

To be copied by Webpack, assets must be referenced by modules that use them. If you require / point to ressource in your module, webpack will fid it locally and pipe it to the output dir.

Some of our assets aren't referenced (favicons, mostly). But we need to copy them too. So we use 2 `assets` folders: one in `app`, for assets referenced by modules ; and one in `vendor` which is just copied to the output dir (using the _CopyPlugin_). We can't mix the two folder or it'll be a mess for us and Webpack to know which assets are alreday referenced to not be copied a second time. 

I [first suggest](https://github.com/cozy/cozy-proxy/pull/246#issuecomment-187261432) to have a public folder at root level, but it brings front-end related assets outside of the client directory.

m4dz:
> Could you make the file < 80 chars wide?

Done :smiley:

frankrousseau:
Thank you @m4dz. The Cozy Proxy client folder looks a lot cleaner now.

PR #266, opened 2016-04-22
Fix locales support

m4dz:
This PR:
- Remove the unneeded `defaultPolyglot` instance in localization manager (Transifex serves untranslated string by default, non need to do it manually)
- compute the list of available fonts directly for the _locales_ dir
- fix the returned `lang` attribute that was corrupted


nono:
Thanks!

PR #294, opened 2016-07-15
Fixed the way the 2FA key is being handled.

babolivier:
To publish at the same time than https://github.com/cozy/cozy-home/pull/771.

Fixed the way the 2FA master key is being handled by Cozy's proxy, so a Yubikey can be used to authenticate through Cozy.

Also removed an useless update which could lead to master key corruption if handled wrongly.

### DIFF
--- a/server/lib/passport_configurator.coffee
+++ b/server/lib/passport_configurator.coffee
@@ -43,11 +43,10 @@ module.exports = ->
     # hotp strategy
     # we need a key (stored string) and a counter
     passport.use new HotpStrategy { codeField: "authcode" }, (user, done) ->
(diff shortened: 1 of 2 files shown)


aenario:
Hey @babolivier, could you check why Travis got red (https://travis-ci.org/cozy/cozy-proxy/jobs/145027655#L1304-L1345) ?

babolivier:
Oops, seems like I forgot to change the test. On it as soon as I get home. Thanks for noticing @aenario.

babolivier:
@aenario Fixed the failed test, all is green now.

nono:
Nice!

PR #300, opened 2016-09-28
Onboarding `view` documentation 

misstick:
# Onboarding view documentation 

 - [x] explain how to create a step,
 - [x] explain how to add new methods.


doc/state-component/api.md
```diff
@@ -0,0 +1,123 @@
+
+Questions :
```
misstick: @CPatchane, @gregorylegarec, @m4dz: that's for you!

misstick:
Are you expecting other informations from API? Did I miss something big?

doc/state-component/api.md
gregorylegarec: Seems ok for me, except that in my opinion our states should not be Views but some kind of data containers, defined in a agnostic way in a specific file. Those data container could then be used into Backbone Views, but also another framework in the future.

What do you thing @misstick @m4dz @CPatchane ?

misstick: Could you draw that into a schema; I will FIX this after?

misstick: @gregorylegarec : what about `propTypes`; @m4dz would you know another micro framework to handle this?

gregorylegarec:
What about changing step 2 name from `share` to `agreement` ?
 @misstick  @CPatchane @m4dz

misstick:
@gregorylegarec, @m4dz, @CPatchane  : `agreement` is fine but you give an agreement for something; this "something" is missing. Maybe : `share_agreement`?

doc/onboarding/model/user.md
```diff
@@ -0,0 +1,22 @@
...
+
+
+## Definition
+```
+{
+	<string> name,
+	<id> id,
+	<email> email,
+	<boolean> is_registered: false
+	<boolean> is_shared: false
```
CPatchane: After reflexions, using flag is not the best solution to know if the onboarding should be launched or not. I think it could be better to store an array of steps to be submitted by the user. By default this array contains all steps like:

```
[<string>] onboarding_steps: ['welcome',  'agreement', 'password', 'confirmation']
```

If we want the user to resubmit agreements for exemple, we have just to had this step in this array.
Of course, the order of all steps should be stored in the proxy.

What do you think about?

doc/onboarding/view/step1.md
```diff
@@ -0,0 +1,59 @@
...
+<pre>
+user = {
+	&lt;string&gt; name,
+	&lt;id&gt; id,
+	&lt;email&gt; email,
+	&lt;boolean&gt; is_registered: false
+}
+</pre>
+<pre>
+step = {
```
gregorylegarec: My vision was that the step here should be a Backbone `StepModel`, which encapsulate our onboarding `Step` object.

* Backbone `StepModel` because we are in a Backbone/Marionette application and we should keep the most part of its logic
* Our onboarding `Step` object because it allow us to decentralize the business logic and stay framewok-agnostic

doc/state_api.md
```diff
@@ -0,0 +1,57 @@
+# State
```
gregorylegarec: This concern is pretty much handled by the `Onboarding` component.

doc/views_api.md
```diff
@@ -0,0 +1,100 @@
+
+## `StateComponent` API
+
+<pre>
+&lt;class&gt; StateComponent extends Backbone.BaseView
+
+    state: new ProxyState()
```
gregorylegarec: Backbone view are more designed to work with Backbone Models, here we can use `StepModel`.

doc/views_api.md
```diff
@@ -0,0 +1,100 @@
+
+## `StateComponent` API
+
+<pre>
+&lt;class&gt; StateComponent extends Backbone.BaseView
```
gregorylegarec: Actually `StateComponent` seems to be an abstract class for all `StepViews`, what do you think ?

doc/views_api.md
```diff
@@ -0,0 +1,100 @@
+
+## `StateComponent` API
```
CPatchane: This is a ```StepComponent```. I think that there is kind of confusion between ```step``` and ```state```

doc/onboarding/model/step.md
```diff
@@ -0,0 +1,22 @@
+# `step` State
```
CPatchane: It seems to be specific of the ```onboarding``` component which is not a step.

doc/onboarding/model/form.md
```diff
@@ -0,0 +1,30 @@
+# `form` State
+
+## Definition
+State of a form that includes its values but also :
+ - required field or
+ - ability to submit or not all the form.
+
+```
+{
+    <boolean> disabled: false
```
CPatchane: Is there a user case when we need a form that is submittable with errors? I only see the case when a step is not mandatory and the user can skip it. If so, I think it's better to use ```skippable``` or ```mandatory``` as boolean here.

gregorylegarec: At this time do we have a not-mandatory step ? If not, maybe it's not our concern for the moment.

doc/onboarding/view/step2.md
```diff
@@ -0,0 +1,82 @@
+
+
+## Step 2/5 : Share anonymously data
+
+### URI
+
+`:userID/share`
```
misstick: @gregorylegarec I thougth you read the comment bellow (that I'v juste removed not to have twice the same question); here it is : 

@gregorylegarec, @m4dz, @CPatchane : agreement is fine but you give an agreement for something; this "something" is missing. Maybe : share_agreement?

doc/onboarding/model/form.md
misstick: @gregorylegarec, @CPatchane : l think that form is disabled when password complexity is not "green"; we have to check with @jsilvestre

doc/onboarding/view/step2.md
CPatchane: It's not only about sharing here, isn't it? It could be better ```agreements``` but I think ```share_agreement``` is not enough explicit here.

doc/onboarding/model/form.md
CPatchane: Ok so it's confused here. A form can not be ```disabled```, that doesn't make sense. If the password is not enough complex and we have an error, it's just ```not validated```. All these stuffs should be handled by the view specific validating function (which is used by the ```onboarding``` component). So maybe this boolean is not necessary. What do you think?

doc/onboarding/view/step2.md
misstick: see https://github.com/cozy/cozy-proxy/pull/300/commits/937ff9041a3b97154b8c3e8191a97d290c81c240

doc/onboarding/model/form.md
misstick: I asked to @jsilvestre and @poupotte : it seems that this props is useless which means that form could be submitted be server would send an error if passwordValue will be too weak.

doc/onboarding/view/step2.md
CPatchane: @misstick Unless I am mistaken, the commit https://github.com/cozy/cozy-proxy/pull/300/commits/937ff9041a3b97154b8c3e8191a97d290c81c240 that you just gave doesn't solve the naming issue here.

doc/onboarding/view/step3.md
```diff
@@ -0,0 +1,106 @@
...
+	&lt;step&gt; value: 3/5,
+}
+</pre>
+
+
+### Getters
+
+<pre>
+    getPasswordComplexity: (state) ->
+        // Apply algorithm to know
```
CPatchane: Agree with @gregorylegarec. The function to get/check the password complexity should be a part of the password component, not the view. The view will check the password complexity using this function to validate the form/view.

doc/onboarding/view/step1.md
misstick: see https://github.com/cozy/cozy-proxy/pull/300/commits/d3f891c60a7933c295cf44791fcbd0530d206cb8

doc/onboarding/view/step3.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Actions
+
+<pre>
+    getFormData: ->
+        return { password: @state.password, id: @state.userID }
+
+
+    validate: ->
+        { password } = @getFormData()
```
CPatchane: @gregorylegarec I am not sure. For me, the onBoarding component shouldn't be aware about form data, like password here. So we have to use a validate function without arguments and grab form data inside its execution in the view context. What do you think?

doc/onboarding/view/step1.md
```diff
@@ -0,0 +1,59 @@
...
+	&lt;string&gt; slug: 'welcome',
+	&lt;step&gt; value: 1/5
+}
+</pre>
+
+### Getters
+
+<pre>
+    getNextStepURI: (state) ->
```
misstick: A `getter` is a specific object that doesnt belong to views or models. Views get data from this object that will get data from models so that views never calls models directly (immutability). That's a pattern used with React and Redux framework.

I d'on thing that you implemented such a thing; my question is : should we create one? Where should it be?

CPatchane: I think here is just about getting the next step URI. @gregorylegarec said that the view shouldn't know the next step, all the routing is handled by the ```onboarding``` component. If the user want to go to the next step, there is just a function ```goNext()``` that use the ```onboarding``` component. This latter will run the validating function like to submit a form and redirect to the next step, from outside the current view.

It's just about details but I disagree about immutability here @misstick. For me, immutability is not about using ```getter````/```setter```, it's about using only unchangeable state object. We could also use ```getter``` with mutable object.

doc/onboarding/view/step3.md
gregorylegarec: Onboarding, no, but Step, I think yes. For me a step is aware of how to validate itself. Maybe my example was bad : the onboarding should not know that a given step needs explicit `password` to validate itself. Maybe the call should look like:
```javascript
@step.validate()
// or
@step.validate(@step.getData())
```
Was it your point ?

doc/onboarding/model/step.md
gregorylegarec: @CPatchane  This morning I changed my mind about this one. Because what we need in the view is : current step index and total steps. 
Current step index is related to the current step.
Total steps is related to the onboarding and will not change during user's course.

doc/onboarding/view/step3.md
CPatchane: You're right, I agree with this last point that answers my question here. All data are grabbed inside the validation function and not provided by the onboarding component but by the view/step. Thanks.

misstick:
@gregorylegarec : can you review this file which is the only document that still exist in this PR? https://github.com/cozy/cozy-proxy/blob/documentation/state-component-API/doc/onboarding/views.md

m4dz:
Great, thanks!

PR #303, opened 2016-10-03
Add onboarding component

gregorylegarec:
Contains class, test, and doc. Integration will follow.


client/app/lib/onboarding.coffee
```diff
@@ -0,0 +1,111 @@
+# Local class Step
+class Step
+    # Retrieves properties from config Step plain object
+    constructor: (step) ->
+        if step
```
misstick: Why not use this instead?
````
constructor: (step={}) ->
 unless _.isEmpty step 
# then do ...

````

File: client/app/lib/onboarding.coffee [hunk 2]
```diff
@@ -0,0 +1,111 @@
...
+    onValidated: (callback) ->
+        throw new Error 'Callback parameter should be a function' \
+            unless typeof callback is 'function'
+        @validatedHandlers = @validatedHandlers or []
+        @validatedHandlers.push callback
+
+
+    # Trigger 'validated' pseudo-event
+    triggerValidated: () ->
+        if @validatedHandlers
```
misstick: @gregorylegarec : I find a contradiction between this line and line 14 above. Furthermore I don't find any default value such as : 
`````
class Step

    validatedHandlers: null
````

File: client/app/lib/onboarding.coffee [hunk 3]
```diff
@@ -0,0 +1,111 @@
...
+                @handleStepSubmitted step
+            return stepModel
+
+
+    # Records handler for 'stepChanged' pseudo-event, triggered when
+    # the internal current step in onboarding has changed.
+    onStepChanged: (callback) ->
+        throw new Error 'Callback parameter should be a function' \
+            unless typeof callback is 'function'
+        @stepChangedHandlers = @stepChangedHandlers or []
```
misstick: Why not add a method for this (ever called line 14) such as : 
`````
addCallback: (callback) ->
    @stepChangedHandlers ?= []
    @stepChangedHandlers.push callback
````

File: client/app/lib/onboarding.coffee [hunk 4]
```diff
@@ -0,0 +1,111 @@
...
+        @stepChangedHandlers.push callback
+
+
+    # Handler for 'stepSubmitted' pseudo-event, triggered by a step
+    # when it has been successfully submitted
+    # Maybe validation should be called here
+    # Maybe we will return a Promise or call some callbacks in the future.
+    handleStepSubmitted: (step) ->
+        @goNext()
```
misstick: Why defining `step` as arguments if it is no used?

client/app/lib/onboarding.coffee
```diff
@@ -0,0 +1,111 @@
+# Local class Step
+class Step
```
CPatchane: It could be better to put this 'Step' class in a separate step.coffee file, isn't it?

client/app/lib/onboarding.coffee
```diff
@@ -0,0 +1,111 @@
...
+
+
+    initialize: (user, steps) ->
+        throw new Error 'Missing mandatory `steps` parameter' unless steps
+
+        @user = user
+        @steps = steps.map (step) =>
+            stepModel = new Step step
+            stepModel.onValidated (step) =>
+                @handleStepSubmitted step
```
misstick: Why not do this instead (and remove line50)
`````
stepModel = new Step step, {
    submit: @handleStepSubmitted
}

class Step = 
    # ...
    submit: (args...) ->
         if (isValidated = @validate args...)
             @triggerValidated()
             @callback.submit args...
`````

File: client/app/lib/onboarding.coffee (hunk 1 above)
Comment (CPatchane) [inline client/app/lib/onboarding.coffee]: If we could keep this as generic as possible and with the minimum dependancies, I think it's better for the reusability. I don't see a big interest to use underscore here.

File: client/app/lib/onboarding.coffee [hunk 7]
```diff
@@ -0,0 +1,111 @@
...
+    handleStepSubmitted: (step) ->
+        @goNext()
+
+
+    # Go to the next step in the list.
+    goNext: () ->
+        currentIndex = @steps.indexOf(@currentStep)
+
+        if @currentStep? and currentIndex is -1
+            throw Error 'Current step cannot be found in steps list'
```
misstick: Don't you need to use @currentStep for error reporting?

client/app/lib/onboarding.coffee
CPatchane: If we could keep this as generic as possible and with the minimum dependancies, I think it's better for the reusability. I don't see a big interest to use underscore here.

client/app/lib/onboarding.coffee
```diff
@@ -0,0 +1,111 @@
...
+        nextIndex = currentIndex+1
+
+        if @steps[nextIndex]
+            @goToStep @steps[nextIndex]
+        else
+            @triggerDone()
+
+
+    # Go directly to a given step.
+    goToStep: (step) ->
```
misstick: I think that `gotoStep` is more right in camelCase use.

CPatchane: I personally prefer the @gregorylegarec naming here. To discuss.

client/doc/onboarding.md
```diff
@@ -0,0 +1,64 @@
+## Onboarding component
```
misstick: @gregorylegarec : Sorry but I d'ont see the point of this file; it seems to me that it is overide `doc/api`.

CPatchane: To my mind, this is the description/doc of the onboarding machine, not the doc of the component that will be handled by this onboarding object (which is your PR).

client/app/lib/onboarding.coffee
CPatchane: I personally don't see the interest here while it's used only once in each class here.

client/app/lib/onboarding.coffee
misstick: The point is not about underscore; the point is about initialize step with the expected type.

client/app/lib/onboarding.coffee
misstick: You're totally right; I was sure to have seen this twice. My bad!

client/doc/onboarding.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Onboarding class
+
+This class, located in `lib/onboarding`, is an agnostic and framework free onboarding manager. It manages the different steps of onboarding and handle event triggered by steps object.
+
+The goal of this class is facilitate migration to another framework in the future.
```
misstick: Maybe start with the goal before explaining how it works?

client/doc/onboarding.md
```diff
@@ -0,0 +1,64 @@
...
+
+The class represents an onboarding step, for example the greetings step or the password definition step.
+
+It's in charge of describing what a step have to do and how it may do that.
+The main idea is to be able to describe what the onboarding steps are showing/containing in external files, and to easily integrates them in the onboarding engine.
+
+## Methods
+### Onboarding
+#### initialize(user, steps)
+Set the user and the steps list for the current onboarding. Called by the constructor method. The steps list should look like :
```
misstick: Small sentences; would this be more readable if each is on a new line?

client/doc/onboarding.md
```diff
@@ -0,0 +1,64 @@
...
+#### onStepChanged(callback)
+##### Parameters
+* `callback`: function
+
+Record the function callback as handler for every time the current onboarding step will change.
+
+#### goNext()
+Select the next step on the list and trigger the related events.
+
+#### goToStep(step)
```
misstick: Maybe you should choose between `goto` and `go` (see line 40 & 43)

client/doc/onboarding.md
```diff
@@ -0,0 +1,64 @@
...
+Go directly to the given step and trigger required events. Useful for go back in onboarding.
+
+#### getStepByName(stepName)
+##### Parameters
+* `stepName`: String
+
+Returns a step by its name.
+
+### Step
+#### onSumbitted(callback)
```
misstick: typo!

client/doc/onboarding.md
```diff
@@ -0,0 +1,64 @@
...
+Select the next step on the list and trigger the related events.
+
+#### goToStep(step)
+##### Parameters
+* `step`: Step
+
+Go directly to the given step and trigger required events. Useful for go back in onboarding.
+
+#### getStepByName(stepName)
+##### Parameters
```
misstick: Dont we use `arguments` instead of `parameters` ?

client/app/lib/onboarding.coffee
gregorylegarec: It's more Vanilla JS economy habit, Coffee will be transpiled with more lines of code to implements this feature. But ok let's use it.

client/app/lib/onboarding.coffee
gregorylegarec: Two differents visions. If you want.

client/app/lib/onboarding.coffee
gregorylegarec: I feel like it will be useful in the future, but you're right, let's avoid dead code.

client/app/lib/onboarding.coffee
gregorylegarec: In my opinion, `StepModel` is enough correlated to `Step` to avoid adding complexity by passing callback and handlers as constructor parameters.

client/app/lib/onboarding.coffee
gregorylegarec: My opinion is that Step is strongly linked to Onboarding, and almost exclusively used by it. It is almost an internal class.

Furthermore, onboarding cannot be required without Step with this pattern.

But if you prefer I may externalize it.

client/app/lib/onboarding.coffee
gregorylegarec: Honestly I don't really care, but I feel that the rule one word/one capital make more sense.

client/doc/onboarding.md
gregorylegarec: Yep you're right.

client/doc/onboarding.md
gregorylegarec: You're absolutely right, thanks.

client/doc/onboarding.md
gregorylegarec: Nope :)
http://stackoverflow.com/questions/1788923/parameter-vs-argument

client/doc/onboarding.md
gregorylegarec: Thanks !

client/app/lib/onboarding.coffee
gregorylegarec: I fixed as required but I changed my mind about this one :
* First, if we declare it as a class property it will be a global static variable, shared by  all instances
* Second, another way is to declare `@validateHandlers` as an empty array in `initialize`, but I am not comfortable with the way of always declaring an empty array for every instance. I prefer to instanciate it on demand in `onStepValidated` and check existence in `triggerValidated`. Even if in fact and in our code, Step will always have handlers in `validatedHandlers`. So if you don't mind I will let this part as it.

client/app/lib/onboarding.coffee
gregorylegarec: Ok :) we may let it as it right now and externalize it in the future if needed.

client/app/lib/onboarding.coffee
misstick: @gregorylegarec : "think wise not too much"; 👍

package.json
```diff
@@ -75,9 +75,10 @@
     "postbuild": "npm-run-all --parallel copy:*",
     "postinstall": "node postinstall.js",
     "start": "node build/server.js",
-    "test": "npm run test:server",
+    "test": "npm run test:server; npm run test:client",
```
m4dz: Details, but you can use `npm-run-all 'test:*'" to serialize the run of all `test:`scripts #protip

nono: At least, you should do `npm run test:server && npm run test:client`. Else, if the `npm run test:server` fails but not the `npm run test:client`, the return code will still be OK and the failure can be easily missed.

gregorylegarec: Ok thanks guys.

Let's go for `npm-run-all 'test:*'"`

gregorylegarec: @m4dz version recursively calls `"test:build": "env USE_JS=true npm run test"`, do I delete this command, rename it or use @nono's solution ?

nono: I think you can delete this command. It looks like an old command that does no longer do what you will expect to do (testing the built JS files).

gregorylegarec: Done :)

CPatchane:
Thans @gregorylegarec

PR #304, opened 2016-10-04
Integrate onboarding component into existing application

gregorylegarec:
__Based on PR #303__

The main goal of this PR is to integrate previously commited onboarding into existing proxy application.

## How it works

The idea is to get standalone and separated objects defining onboarding steps. They are located into the `steps` directory and listed into `steps/config.coffee`.

The application requires those steps objects and pass them to the onboarding component.

Steps are minimal at this time, but we will enhance them in the next developments, and add methods dealing with validation logic, data fetching, etc. (if needed).

After that, the application set up the router with specific routes for the steps. This part has been pretty poorly tested and it will surely need adjustment in the next improvements.

The main ideas are :
* Set back the logic into the application class, to get all business code located in one place.
* Configure and use the router externally, as his role should only be to route URL to methods.

All those points are of course discussable. However my opinion is that we should use it as it for now and implements one change at a time in the next developments.

## Modelisation naive diagram
Diagram

### StepModel
The StepModel class may be a seen as a useless layer, but its role is to wrap our agnostic Step object to Backbone Model. So we can use it easily in Marionette View.

## Next Steps
* Our next work will be to plug it out with @misstick works on steps components API (or views); see https://github.com/cozy/cozy-proxy/pull/300

* When all onboarding steps are done, nothing is implemented for the moment.

__All this points are of course disccussable__.


misstick:
@gregorylegarec : `steps objects` is it data, view, controller?

client/app/application.coffee
```diff
@@ -25,8 +27,13 @@ class App extends Application
     - layout: the application layout view, rendered.
     ###
     initialize: ->
+        steps = require './steps/config'
         @on 'start', (options) =>
-            @router = new Router app: @
+
+            @onboarding = new Onboarding({}, steps)
+            @onboarding.onStepChanged (step) => @handleStepChanged(step)
```
misstick: @gregorylegarec : couldnt you use `baconjs` to listen to data? If not could you explain it simply (with an example?)

client/app/application.coffee
```diff
@@ -37,5 +44,44 @@ class App extends Application
             Object.freeze @ if typeof Object.freeze is 'function'
 
 
+    # Initialize routes relative to onboarding step.
+    # The idea is to configure the router externally as a "native"
+    # Backbone Router
+    # @param steps a list of Step instance
+    initializeRouter: (steps) ->
```
misstick: @gregorylegarec : why not move this into a custom `Router` object?

client/app/models/step.coffee
```diff
@@ -0,0 +1,17 @@
+Backbone = require 'backbone'
+
+# StepModel
+# Backbone wrapper for Onboarding's Step models
+# This Class is a proxy between Backone's logic and native Onboarding Steps
+module.exports = class StepModel extends Backbone.Model
+
+    # Map needed property to current model
+    initialize: ({step}) ->
```
misstick: I don't understand what the nature of `step` in argument; could you comment this part plz? thx :)

client/app/steps/config.coffee
```diff
@@ -0,0 +1,14 @@
+# List of onboarding steps.
+# Every steps are in the same file now, but the idea is to
+# At this time there is only proper
+welcome = require './welcome'
+agreement = require './agreement'
```
misstick: See my comment on https://github.com/cozy/cozy-proxy/pull/300

client/doc/onboarding.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Onboarding class
+
+This class, located in `lib/onboarding`, is an agnostic and framework free onboarding manager. It manages the different steps of onboarding and handle event triggered by steps object.
+
+The goal of this class is facilitate migration to another framework in the future.
+
+At this time, the framework used is Backbone/Marionette and Onboarding is implemented like a classical POO object.
+
+## Step class
```
misstick: I dont understand nature of this class; is this the state of the application? is this a model?

misstick:
@gregorylegarec : could add a schema to explain briefly relations between components? You are adding a lot of notions that are not necessarily obvious in one look; and the documentation isnt explaining me much about that :(

Otherwise : great "big" work :)

client/app/application.coffee
gregorylegarec: Here we are just dealing with native onboarding component, that we designed framework and lib agnostic. Such a simple thing as dealing with one or two events should not require a library in my opinion. The code is flexible enough to evolve to a library dependant logic in the future.

client/app/models/step.coffee
gregorylegarec: You're right it is too poorly commented.

gregorylegarec: I am thinking about renaming Step to OnboardingStep in the future, to clarify. Let's see that after the merge of this PR and the previous one. I am interested by your opinion about this point @CPatchane and @m4dz. :)

client/doc/onboarding.md
gregorylegarec: Its an object, in a purely POO paradigm. It contains only properties at this time but it will contains other needed methods in the future. I think for example that validation logic should be implemented here and just called from the views. I clarifying the documentaiton on this point.

client/app/models/step.coffee
m4dz: I'm not really sure about what you want to rename here…?

gregorylegarec: @m4dz The `Step` class, used in `Onboarding` class, to more explicit name `OnboardingStep`.

m4dz: I though the `Step` class is an internal class for `Onboarding`, and isn't exposed outside of it? Is it not the case now?

CPatchane: I think if the ```Step``` class is an exclusive class of ```Onboarding``` we can keep it like that. But if this should be an extern class that should be renamed.

misstick:
@gregorylegarec : I don't see relation between `collection/models` and application `controller`to get data (missing `feed arrow`beside `listen` arrow?

misstick:
@gregorylegarec : about `step/` directory; shouldnt it be renamed into `config/` and then be consistent with your schema?

gregorylegarec:
As `Application` is listenning to `Onboarding`, to me it implies that it has access to onboarding's data/info. That also why I called this diagram "naive", only main relationships are mentionned.

gregorylegarec:
@about I prefer `steps` directory, or at least `config/steps` but `config` is too generic. If you're ok with `config/steps` I may update this, it clarify a little bit the role of those files.

gregorylegarec:
@misstick I just renamed it.

CPatchane:
Great! Thanks @gregorylegarec

PR #305, opened 2016-10-06
Onboarding Welcome step - markup

m4dz:
This PR introduce markup and style for the first step (welcome) of new onboarding.

- [x] static markup and poc style
- [ ] report static markup in app views
- [ ] create dedicated stylesheet for this step (related to #315 )


static/index.html
```diff
@@ -0,0 +1,45 @@
...
+    <header>
+        <a href="https://cozy.io" class="cozy">
+          <img src="assets/illustration-welcome.svg" alt="Welcome to Cozy cloud" />
+        </a>
+
+        <h1>Welcome Michel</h1>
+        <a href="#">You're not Michel?</a>
+    </header>
+
+    <div role="region">
```
nono: For my culture, what is the interest of this role compared to a simple `<section>`?

static/index.html
```diff
@@ -0,0 +1,45 @@
...
+      <p>
+        Congratulation, thanks to you, your personal data are in a safe place, in your Cozy.
+      </p>
+    </div>
+
+    <footer>
+      <div class="controls">
+        <a href="#" role="button">next</a>
```
nono: By curiosity, why do use a link with `role="button"` instead of just a `<button>`? Is it because it's easier to style and to avoid browser specific issues?

static/main.styl
```diff
@@ -0,0 +1,75 @@
...
+        &:last-child
+            margin-right: 0
+
+        &[aria-selected]
+            background-color: $blue
+
+
+[role=region]
+footer
+    font-size: 1.24em
```
nono: You don't use rem for font-size?

static/index.html
m4dz: Good question: a `[role=region]` describes

> A large perceivable section of a web page or document, […] for example, an area of the page containing live sporting event statistics.

(https://www.w3.org/TR/wai-aria/roles#region)

In fact, it is similar to a section, but AT aims to observes a region to render quickly when they're updated, so it seems to be a good candidate, as this part of the onboarding will be very dynamic.

static/index.html
m4dz: I use it here because I'm not sure at this time how the next button will be handled in code. As steps have their own URIs, and as long the welcome screen doesn't contain a <form> to submit, this one is probably more a link to the next step rather than a button that doesn't send anything.

static/main.styl
m4dz: I prefer to rely on rem for layout sizes (doxes widht/height), and use relative font-sizes for typography. But it's just a matter of taste, not a real recommandation in this context.

In a more strictly CSS framework, I'll probably force the use of `em` for typography as long as they're all relatives to each others in the DOM tree representation.

m4dz:
> Just a little doubt about the CSS class name "cozy"

Yeah, me too. Ill try to find something more explicit (dunno at this time)

m4dz:
Oops, just seen this one is comming for the wrong branch in my repo. Closing and reopen it soon.

PR #307, opened 2016-10-11
Add active steps detection

gregorylegarec:
Thanks @m4dz to have a review. (ping also @CPatchane and @misstick)

This PR add a way to detect if an onboarding step is active, i.e. it it has to appear in the onboarding flow.

To handle that, it adds an `isActive(user)` method to Onboarding Step class, which specifies if the step has to be active for a given user (See documentation).

This method may be overriden by config steps. See for example infos step.

### Infos step

Also, it add the infos step, only active when given user don't have valid `email` or `timezone`.

### Also added
* Documentation
* Tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,6 @@ before_install:
 
 script:
     - npm run build
-    - npm run test:build
 
 after_failure:
     - pwd
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,7 @@ before_install:
 
 script:
     - npm run build
+    - npm run test
 
 after_failure:
     - pwd
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ branches:
   except:
     - demo
 node_js:
-    - 0.10
-    - 0.12
     - 4
     - 5
 services:
--- a/client/app/lib/onboarding.coffee
+++ b/client/app/lib/onboarding.coffee
@@ -30,7 +30,7 @@ class Step
     # @param user : plain JS object. Not used in this abstract default method
     #  but should be in overriding ones.
     isActive: (user) ->
-        return true;
+        return true
 
     # Submit the step
     # This method should be overriden by step given as parameter to add
--- a/client/app/lib/onboarding.coffee
+++ b/client/app/lib/onboarding.coffee
@@ -1,11 +1,12 @@
 # Local class Step
 class Step
     # Retrieves properties from config Step plain object
+    # @param step : config step, i.e. plain object containing custom properties
+    #   and methods.
     constructor: (step={}) ->
-        @validatedHandlers = []
-
-        ['name', 'route', 'view'].forEach (property) =>
-            @[property] = step[property]
+        ['name', 'route', 'view', 'isActive'].forEach (property) =>
+            if step[property]
+                @[property] = step[property]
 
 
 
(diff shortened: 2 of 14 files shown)


nono:
@gregorylegarec coffeelint is not happy :wink:

gregorylegarec:
@nono Oops, sorry. It's the second time that it happens to me, I will see how I could add coffeelint to the npm watch task. :)

gregorylegarec:
@nono Should be fixed now.

client/doc/onboarding.md
```diff
@@ -61,6 +61,30 @@ Go directly to the given step and trigger required events. Useful for go back in
 Returns a step by its name.
 
 ### Step
+
+#### isActive(user)
+* `user`: JS Object
+
+Returns true if the step has to be active for the given `user`. Returns `true` by default.
```
misstick: I'm not sure that your first sentence his usefull. It seems obvious according to the method name. Furthermore your exemple below speaks for itself.
Precise default value by the way is great 👍

client/doc/onboarding.md
```diff
@@ -61,6 +61,30 @@ Go directly to the given step and trigger required events. Useful for go back in
...
+This method can be overriden by specifying an `isActive` method in constructor parameter.
+
+##### Example
+```javascript
+let step = new Step({
```
misstick: Could you precize where should be this code? Where is it called on this project? Is it Application.initialize()?

nono:
@gregorylegarec you can also remove node 0.10 and node 0.12 in `travis.yml`

.travis.yml
```diff
@@ -8,8 +8,6 @@ branches:
   except:
     - demo
 node_js:
-    - 0.10
-    - 0.12
     - 4
     - 5
```
gregorylegarec: Could we make a dedicated PR for this stuff ? We are starting to mix to concerns here :)

client/app/lib/onboarding.coffee
```diff
@@ -46,11 +54,15 @@ module.exports = class Onboarding
...
-            stepModel.onValidated () =>
-                @handleStepSubmitted()
-            return stepModel
+        @steps = steps
+            .reduce (activeSteps, step) =>
+                stepModel = new Step step
+                if stepModel.isActive user
+                    activeSteps.push stepModel
+                    stepModel.onValidated () =>
+                        @handleStepSubmitted()
```
gregorylegarec: Good question, I asked it to myself. Actually if we do this like it, we lose the onboading `this` context, as CoffeeScript does not offer a way to directly bind a function (or maybe I just don't know how to do). Wrapping this code in another function is the only way I found keep this context.

.travis.yml
m4dz: Al'right, I agree 👍

client/app/lib/onboarding.coffee
m4dz: Hmm… we can declare in https://github.com/cozy/cozy-proxy/pull/307/files#diff-753b070baa8f3302572f4381d40ac8daR80 

```coffee
handleStepSubmitted: =>
    @goToNext()
```

But if I understand code properly, this method can be overriden in Step, so there's a risk to loose the binding, right?

If this, just take it as it :)

gregorylegarec: For the moment, this method is not designed to be overriden. So let use your solution, thanks 👍

m4dz:
Thanks @gregorylegarec !

Just for information, we'll try to not forget to update localization keys in steps markups 😉

PR #308, opened 2016-10-11
[password/step1] Add step password input

misstick:
## Tasks
 - [x] Add basic input into markup,
 - handle submit case:
  - [x] handle `validate`,
  - [x] handle `passwordValue` registration.

## Tests
 - [x] Password specific callback (`submit`& `validate`),
 - [x] `PasswordSubmit` must be called before `defaultSubmit` (do not override),
 - [x] `PasswordSubmit` must redirect to `nextStep`.


misstick:
As my dev environment is buggy, I can't test this.

client/app/views/steps/password.coffee
```diff
@@ -1,14 +1,27 @@
 StepView = require '../step'
+timezones = require '../../lib/timezones'
```
CPatchane: Timezones should not be used in the password coffee, it's not a necessary data for this component because it's only password concerned.

client/app/config/steps/password.coffee
```diff
@@ -2,4 +2,16 @@ module.exports = {
...
     view : 'steps/password'
+
+    # If OK, return null
+    # if not return an Array of errors
+    validate: (data) ->
+        return null
+
+    submit: (data) ->
+        jQuery = global.jQuery
+        jQuery ?= require 'jQuery'
```
CPatchane: For my culture, why use two potentially different sources for the jQuery dependance here? Is it not a potential source of issues?

client/app/views/steps/password.coffee
```diff
@@ -1,14 +1,27 @@
...
+    getDataFromDOM: ->
+        # TODO: get user info
+        # from server
+        return {
+            email: 'toto@cozycloud.cc'
+            public_name: 'toto'
+            timezone: timezones[0]
+            password: @$('input[name=password]').val()
+            allow_stats: false
+        }
```
CPatchane: The password component should send only the password from its form/input (not prefilled). So only two properties should be in the request body here: ```password``` and ```onboardedSteps``` (cf doc [here](https://github.com/cozy/cozy-proxy/blob/development/client/doc/onboardingDataHandling.md#views-data-behaviour-client-side)) like:
```coffee
{ 
    password: @$('input[name=password]').val(),
    onboardedSteps: ['welcome', 'agreement', 'password']
}
```

Other data are handled by other steps separately from this component.

client/app/views/steps/password.coffee
misstick: I totally agrre with you, but its a temporary data will be replaced into https://github.com/cozy/cozy-proxy/pull/310, see https://github.com/cozy/cozy-proxy/pull/310/files#diff-6487018685f0002494bb58611dd1d9fcR36

client/app/views/steps/password.coffee
misstick: @CPatchane : Sorry, I misunderstood :( I will remove useless params; but anyway I don't understand what `onboardedSteps` is for? It actually works without. Could you explain that to me please?

CPatchane: Yeah sure @misstick .
1. The previous ```POST /register``` worked with these parameters but now, since the PR #322, it should handle ```password``` parameter only. We can check that together if you still have an issue about that.
2. The ```onboardedSteps``` allows to detect steps already done by the user. This array will be saved in the ```User``` document and the user will redirected by the server according of its value. For each step, the "complete steps" are sent through this array. For the component here, that should be ```['welcome', 'agreement', 'password']```. Don't hesitate if you have other questions about that.

misstick:
I should use`fetch` instead of `jQuery.post` (IRL from @m4dz and @CPatchane )

client/app/config/steps/password.coffee
```diff
@@ -1,5 +1,16 @@
...
     name: 'password',
     route: 'password',
     view : 'steps/password'
+
+    # If OK, return null
+    # if not return an Array of errors
+    validate: (data) ->
+        return null
+
+    submit: (data) ->
```
gregorylegarec: I think we should not allow to override `submit` and create another method `save` called by `submit` like it already does here with `validate`. It is a more hookable approach and it's actually what I've done here : https://github.com/cozy/cozy-proxy/pull/324/files#diff-7f3ff7953dafb8bab4ffe2534bbae502R5

It also simplifies the code in onboarding.initialize, as we are not keeping native submit in a closure.

As it is done in another PR, modification must be done in this one or #324. It depends on which one will be merged first.

client/app/lib/onboarding.coffee
```diff
@@ -9,10 +9,22 @@ class Step
...
         ].forEach (property) =>
-            if step[property]
-                @[property] = step[property]
+            if step[property]?
+
+                # Do not override submit
+                # such as @submit that allow to goto next step
+                # @submit mussnt be overidden but @isActive yes
+                if property is 'submit'
+                    nativeCallback = @[property]
```
gregorylegarec: See previous comment, we should avoid to override `submit` method.

client/app/models/step.coffee
```diff
@@ -19,6 +19,11 @@ module.exports = class StepModel extends Backbone.Model
 
         @set 'next', next
 
-    # Encapsulate call to step.submit
-    submit: () ->
-        @step.submit()
+
+    submit: (data) ->
+        # TODO: should handle errors here
+        if @step.validate? and (errors = @step.validate(data))
```
gregorylegarec: Just a thought : maybe `validate` will be asynchronous in the future (checking existing email ?), maybe it will need to return some kind of Promise. But again, let's see later.

client/app/views/steps/password.coffee
```diff
@@ -1,14 +1,24 @@
...
+timezones = require '../../lib/timezones'
+
 
 module.exports = class PasswordView extends StepView
+
     template: require '../templates/view_steps_password'
 
     events:
-        'click button': 'onSubmit'
+        'click button': 'doSubmit'
```
gregorylegarec: Why renaming here ? `onEvent` is a pretty much accepted naming for specifying event handlers (see https://www.reddit.com/r/javascript/comments/1uwjcp/naming_conventions_when_to_use_on_prefix/) and (http://stackoverflow.com/questions/11990176/backbone-js-best-practice-for-event-handlers-naming), but if you prefer to stick more and what has already been done in cozy-proxy, like here https://github.com/cozy/cozy-proxy/blob/development/client/app/views/register/import_google.coffee#L20, i suggest you to use `submit` instead. `doSubmit` is some kind of repetition :) (it is also what backbone documentation shows http://backbonejs.org/#View-events)

Thanks to fix this. :)

client/app/views/steps/password.coffee
```diff
@@ -1,14 +1,24 @@
...
+
+
+    getDataFromDOM: ->
+        # TODO: get user info
+        # from server
+        return {
+            password: @$('input[name=password]').val()
+            onboardedSteps: ['welcome', 'agreement', 'password']
```
gregorylegarec: No need to change but `onboardedSteps` property is not really retrieved from DOM, so it should not be here or the function may be renamed.

client/test/config/steps/password.js
```diff
@@ -0,0 +1,59 @@
...
+describe('Step: password', () => {
+
+    let jQuery;
+    let jsdom;
+    let PasswordConfig;
+    let TimeZones;
+
+
+    before(function () {
+      jsdom = require('jsdom-global')();
```
gregorylegarec: By removing jQuery dependency I hope we will be able to also skip jsDom dependency. :)

client/test/lib/onboarding.js
```diff
@@ -741,15 +742,15 @@ describe('Onboarding.Step', () => {
...
-            });
+            let defaultSpy = sinon.spy(Step.prototype, "isActive");
+            let step = new Step({ isActive: spy });
 
             // act
             let result = step.isActive();
 
             // assert
             assert(spy.calledOnce);
+            assert.equal(0, defaultSpy.callCount);
```
gregorylegarec: This is already tested in test l.753 and should not be here. If you want, you may add a dedicated test for `prototype.isActive`, but test l.753 does pretty much the same.

client/app/config/steps/password.coffee
gregorylegarec: Addendum: let's merge the current PR first

client/app/views/steps/password.coffee
misstick: I prefer use `onSubmit` not to be confused with `OnboardingStep.submit`

client/app/views/steps/password.coffee
misstick: @gregorylegarec I added this because @CPatchane asked for; maybe we need to be consistent/agree about this; what should I do?

gregorylegarec:
Thanks @misstick 👍

PR #310, opened 2016-10-11
[password/step2] Handle Errors into step password

misstick:
## Tasks
 - [x] Add error handling into `onboardingLib` (global behavior),
 - [x] Add error handling into `config/password` (specific behavior) ,
 - [x] add documentation on new onboarding methods,

## Tests
 - [x] handle error with `configPasswordValidate`,
 - [x] handle error with  `configPasswordSave`,


client/test/config/steps/password.js
```diff
@@ -0,0 +1,92 @@
+'use strict';
+let assert = require('chai').assert;
+let sinon = require('sinon');
+
+
+describe('Step: password', () => {
+
+    let jQuery;
+    let jsdom;
```
misstick: TODO: I should mock `configPasswordSubmit` not to call `jQuery.post`; therefor I should remove `jsdom`, and `jQuery` in these tests.

client/app/styles/onboarding.styl
```diff
@@ -33,13 +33,24 @@ header
             display: block
 
         svg
-            max-width: 100%
-            height:    (200 / 16)em
```
misstick: @m4dz : I had to reduce size, but I didnt understand this ratio; could you explain  it to me so that I can fix this?

misstick:
this PR doesnt works anymore; username missing always sent; see when part1 will be merged

client/app/styles/onboarding.styl
misstick: see https://github.com/cozy/cozy-proxy/pull/310/commits/008056f0933782281ef2bcc25e7b6df2954d2345

misstick:
@gregorylegarec : ready to review

client/app/config/steps/password.coffee
```diff
@@ -7,16 +7,23 @@ module.exports = {
...
+        unless data.password?
+            return { type: 'password', text:'step empty fields'}
+        else
+            return null
+
 
     save: (data) ->
-        data = JSON.stringify data
         return new Promise((resolve, reject) ->
             jQuery.post({
```
CPatchane: Suggestion: Would it be possible to use fetch (like here https://github.com/cozy/cozy-proxy/blob/development/client/app/config/steps/welcome.coffee#L6) instead of jQuery.post and maybe remove the jQuery dependance here? That's also good to keep coherence in our code.

misstick: As discussed before with @m4dz, and @gregorylegarec using fetch is an improvement but not the scope of this PR.

PR #311, opened 2016-10-12
Add progression mechanism and progression indicator into views

gregorylegarec:
Will not work without #307. I tried with this PR to push a rebased version of my branch to isolate only commits related to the progression feature.

This PR adds the method `getProgression(step)` to Onboarding class.

It also implements the logic to get it displayed into views.

## Important changes

This PR introduce a StepView "abstract" class. All step views now inherit from this class.

The markup used in views for progression indicator is the same than the proposed one in @m4dz's PR #305. So all step view templates are changing a little bit but it is absolutely not their final version.

Also added :
* Tests
* Documentation

### DIFF
--- a/client/app/lib/onboarding.coffee
+++ b/client/app/lib/onboarding.coffee
@@ -119,5 +119,17 @@ module.exports = class Onboarding
         return @steps.find (step) ->
             return step.name is stepName
 
+
+    # Returns progression associated to the given step object
+    # @param step Step which we want to know the related progression
+    # returns the current index of the step, from 1 to length. 0 if the step
+    # does not exist in the onboarding.
+    getProgression: (step) ->
+        return \
+            current: @steps.indexOf(step)+1,
+            total: @steps.length,
+            labels: @steps.map (step) -> step.name
+
+
 # Step is exposed for test purposes only
 module.exports.Step = Step
--- a/client/app/views/steps/subviews/progression.coffee
+++ b/client/app/views/steps/subviews/progression.coffee
@@ -1,4 +1,4 @@
-{LayoutView} = require 'backbone.marionette'
+{ItemView} = require 'backbone.marionette'
 
-module.exports = class ProgressionView extends LayoutView
+module.exports = class ProgressionView extends ItemView
     template: require '../../templates/progression'
--- a/client/app/views/steps/subviews/progression.coffee
+++ b/client/app/views/steps/subviews/progression.coffee
@@ -1,0 +1,4 @@
+{LayoutView} = require 'backbone.marionette'
+
+module.exports = class ProgressionView extends LayoutView
+    template: require '../../templates/progression'
--- a/client/app/views/templates/view_steps_agreement.jade
+++ b/client/app/views/templates/view_steps_agreement.jade
@@ -2,4 +2,7 @@ extends view_base
 
 block region
     h1=t('Conditions générales')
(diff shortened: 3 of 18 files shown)


m4dz:
@gregorylegarec can you please rebase it on development now #307 is merged in? I'll let the review be more obvous 😄

gregorylegarec:
@m4dz rebased.

client/app/views/steps/subviews/progression.coffee
```diff
@@ -0,0 +1,4 @@
+{LayoutView} = require 'backbone.marionette'
+
+module.exports = class ProgressionView extends LayoutView
```
m4dz: Why using a layoutview instead of a simple Marionette.View here?

client/app/views/templates/view_steps_agreement.jade
```diff
@@ -2,4 +2,7 @@ extends view_base
 
 block region
     h1=t('Conditions générales')
```
m4dz: We must use key strings for localization. Do we doit here, or do your prefer to open another PR to declare those string keys?

client/app/lib/onboarding.coffee
```diff
@@ -119,5 +119,17 @@ module.exports = class Onboarding
...
+
+    # Returns progression associated to the given step object
+    # @param step Step which we want to know the related progression
+    # returns the current index of the step, from 1 to length. 0 if the step
+    # does not exist in the onboarding.
+    getProgression: (step) ->
+        return \
+            current: @steps.indexOf(step)+1,
+            total: @steps.length,
+            labels: @steps.map (step) -> step.name
```
m4dz: I can't figure out why `getProgression` is returning a js object instead of a `ProgressionModel` object. What do I miss here?

client/app/views/steps/subviews/progression.coffee
gregorylegarec: The very first view was a copy-paste from another one, and LayoutView stayed.

I am changing it right now.

gregorylegarec: Ok, actually it's not working with a View, but it does with an ItemView. What do you think is better ?

client/app/views/templates/view_steps_agreement.jade
gregorylegarec: Maybe we could do this during markup integration ? Until now those labels are purely indicative.

client/app/lib/onboarding.coffee
gregorylegarec: Same as Step Object, I try to keep the return values from any Onboarding method as agnostic as possible. ProgressionModel is a pure wrapper for using this result into a Backbone/Marionette View.

client/app/views/steps/subviews/progression.coffee
m4dz: `ItemView` is the regular way offered by Marionette for _simple_ views. So if we want to to it is a proper way, I recommend this one

client/app/views/templates/view_steps_agreement.jade
m4dz: Perfect!

client/app/lib/onboarding.coffee
m4dz: Oh, true. I understand, thanks 😃

m4dz:
Thanks! 🍿

PR #312, opened 2016-10-12
Update Marionnette version to avoid Backbone conflicts (bis)

misstick:
See previous PR : https://github.com/cozy/cozy-proxy/pull/302

You will find history and complete description of the issue here: 
https://github.com/marionettejs/backbone.marionette/issues/2559#issuecomment-106037738

There was missing webpack configuration to link to the right path when calling package `backbone` or `marionette`

### DIFF
--- a/client/webpack.config.js
+++ b/client/webpack.config.js
@@ -133,6 +133,12 @@ module.exports = {
         filename: optimize? 'app.[hash].js' : 'app.js',
         chunkFilename: optimize? 'register.[hash].js' : 'register.js'
     },
+    alias: {
+      // Force all modules to use versions of backbone and underscore defined
+      // in package.json to prevent duplicate dependencies
+      'backbone': path.join(__dirname, 'node_modules', 'backbone', 'backbone.js'),
+      'underscore': path.join(__dirname, 'node_modules', 'underscore', 'underscore.js')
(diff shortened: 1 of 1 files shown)


client/webpack.config.js
```diff
@@ -133,6 +133,12 @@ module.exports = {
         filename: optimize? 'app.[hash].js' : 'app.js',
         chunkFilename: optimize? 'register.[hash].js' : 'register.js'
     },
+    alias: {
+      // Force all modules to use versions of backbone and underscore defined
+      // in package.json to prevent duplicate dependencies
+      'backbone': path.join(__dirname, 'node_modules', 'backbone', 'backbone.js'),
+      'underscore': path.join(__dirname, 'node_modules', 'underscore', 'underscore.js')
```
misstick: As `underscore` is called into to backbone it is a part of the issue.
Furthermore it is the solution proposed in the issue (see PR description)
I think its wise to keep this line.

gregorylegarec: My opinion it to not add useless fixes (if it works, don't fix it), otherwise we may add jquery too, but let's go with it :)
Thanks @misstick 👍

gregorylegarec:
Thanks @misstick

PR #315, opened 2016-10-12
[chore] split stylesheets

m4dz:
This PR introduces a split in stylesheet from thje webpack config. It let us keep old proxy styles and work to new onboarding in parallel (new styles coming from `client/app/styles/onboarding.styl`). Common CSS (like normalize) are output in a dedicated `common.css` target.

### DIFF
--- a/client/app/routes/index.coffee
+++ b/client/app/routes/index.coffee
@@ -35,6 +35,9 @@ module.exports = class Router extends Backbone.Router
     the submitted form.
     ###
     auth: (path, options) ->
+        # Load app stylesheet
+        AppStyles = require '../styles/app.styl'
+
         AuthView  = require '../views/auth'
         AuthModel = require '../states/auth'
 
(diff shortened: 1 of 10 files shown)


client/app/routes/index.coffee
```diff
@@ -74,6 +77,9 @@ module.exports = class Router extends Backbone.Router
     component determined by the step param.
     ###
     register: (step = 'preset') -> require.ensure [], =>
+        # Load onboarding stylesheet
+        AppStyles = require '../styles/onboarding.styl'
```
gregorylegarec: As the register process is not handled by this code anymore, I think that this line should be there https://github.com/cozy/cozy-proxy/blob/development/client/app/application.coffee#L67

Maybe we can create a named handler to clarify this part of `Application`. Something like :

```coffeescript
@router.route "#{step.route}", "route:#{step.route}", () =>
    @handleStepRoute(step)

// [...]
handleStepRoute: (step) ->
    # Load onboarding stylesheet
    AppStyles = require './styles/onboarding.styl'

    # actual code
    @layout.showChildView 'content',
        new StepView
            model: new StepModel step: step 
```

What do you think ?

gregorylegarec:
Thanks @m4dz 👍

PR #317, opened 2016-10-13
[refactor] Rename validated in completed in Onboarding class

gregorylegarec:
The word "validated" was bad and confusing. The real state of the step was rather "completed".

So, I updated some pieces of code and tests to clarify this part.


misstick:
un small step for human a big one for community!

PR #318, opened 2016-10-14
Add getNextStep method to Onboarding

gregorylegarec:
To be able to add a accessible link to next step into views, we need to have a `getNextStep` method on Onboarding class.

This next step is then given to Backbone StepModel as `next` property.

In views, we can access next step's route by using `next.route`.

Thanks @m4dz to have a review.


m4dz:
Woohooo 🎆

PR #319, opened 2016-10-14
Onboarding: welcome step markups

m4dz:
Re-opens #305 from a proper branch.

This PR introduce markup and style for the first step (welcome) of new onboarding.

- [x] static markup and poc style
- [x] report static markup in app views
- [x] create dedicated stylesheet for this step (related to #315 )


client/app/views/app_layout.coffee
```diff
@@ -14,9 +14,9 @@ require 'normalize.css/normalize.css'
 
 module.exports = class AppLayout extends LayoutView
 
-    template: require './templates/layout_app'
+    template: -> '<main/>'
```
gregorylegarec: For my culture : what is it and how does it work ?

gregorylegarec:
Looks great to me !  🎉 
Thanks @m4dz.

client/app/views/app_layout.coffee
m4dz: Backbone's views `template` property expect a function that can be call at rendering, and which returns a string containing the HTML to inject. Here, I've prefer to create a simple function that return my unique tag instead of a separated template file.

Also, I create a card in our technical debt, coz the AppLayout isn't realy relevant anymore and we can drop it soon 😉.

gregorylegarec: Whoo ok, I understand perfectly now. I though it was some kind of magic trick to refer to a 'main' component elsewhere. Let's say it's Friday Fatigue, ok ?

m4dz: Accepted 😝

PR #320, opened 2016-10-14
Retrieve user data for onboarding

gregorylegarec:
It used the the same mechanism as the previous version : pass the username into ENV variable and retrieve in `Application`.

Also, this PR adds the "accounts" step.

(I finally decided to separate this first part from the final PR about welcome step)


CPatchane:
Thanks @gregorylegarec

PR #321, opened 2016-10-17
Add disabled state on next button

m4dz:


misstick:
sorry ...

client/app/styles/onboarding.styl
```diff
@@ -58,15 +59,34 @@ footer
...
+
+    &[aria-busy=true]
+        position: relative
+
+        &::before
+            width:      1em
+            height:     1em
+            @extend     $icon-spinner-white
+            margin-top: -.5em
+            left:       1em
```
gregorylegarec: As seen with @kelukelu : is it possible to put the spinner on the right side ?

client/app/views/steps/welcome.coffee
```diff
@@ -11,9 +25,29 @@ module.exports = class WelcomeView extends StepView
...
+                .prop 'disabled', true
+                .attr 'aria-busy', true
+        else
+            $el.attr
+                'aria-disabled': true
+                'aria-busy':     true
+
+
+    isDisabled: ($el = @ui.next) ->
+        if $el[0].tagName.toLowerCase() in FORMS_DIS_ELEMENTS
```
gregorylegarec: I suggest to not trust the DOM and use a local property here like `@isDisabled`. What do you think ?

m4dz: Great debate :))

client/app/styles/onboarding.styl
gregorylegarec: @kelukelu seemed to be up for a spinner in the right side of the button element. :)

client/app/views/steps/welcome.coffee
m4dz: I understand your argument but I can't figure out if it's really releant or not: using a script like this one, which allaow to submit again, is clearly an volunteer, malintentioned action. I'm not sure if it a good reason to store Ui states in JS instead of using the markup. But your opinion seems reasonnable… I need to think about it a little more…

m4dz:
Done for me 👍

m4dz:
@gregorylegarec ping?

gregorylegarec:
I got to resolve conflicts, so it was not high on my priority list. I hope to merge this one this afternoon.

gregorylegarec:
Thanks @m4dz, especially for your patience here. 👍

m4dz:
Oh, I wasn't in a hurry you kown 👅

PR #309, opened 2016-10-11
[POC/Reflexion] Onboarding data handling using token

CPatchane:
TO NOT MERGE

Add documentation about user data management by the onboarding (.md file). This documentation describe the existing working and data flow.
A proposition is given to:
* [Detect the step according to the user document context](https://github.com/cozy/cozy-proxy/pull/309/files#diff-bafd0c87fdf0cc50bd63327c02fcecc1R75)
* [Describe views behaviours with data](https://github.com/cozy/cozy-proxy/pull/309/files#diff-bafd0c87fdf0cc50bd63327c02fcecc1R90)
* [Give a way to update onboarding steps in the future](https://github.com/cozy/cozy-proxy/pull/309/files#diff-bafd0c87fdf0cc50bd63327c02fcecc1R119)

__Important points:__
* ```welcome```, ```agreement``` and ```password``` steps are processing __before the login__, so the only one user data used is the ```public_name``` [reachable by an environment variable](https://github.com/cozy/cozy-proxy/pull/309/files#diff-d78ee5950212d26245ee1dabefc00efcR11).
* Other steps  will require user authentication to request other user informations from the user document (for pre filled features).

Implementations are still in progress.

PR #322, opened 2016-10-17
Onboarding data handling

CPatchane:
Add documentation about user data management by the onboarding (.md file). This documentation describe the existing working and data flow.
A proposition is given to:
* [Detect the step according to the user document context](https://github.com/cozy/cozy-proxy/pull/309/files#diff-bafd0c87fdf0cc50bd63327c02fcecc1R75)
* [Describe views behaviours with data](https://github.com/cozy/cozy-proxy/pull/309/files#diff-bafd0c87fdf0cc50bd63327c02fcecc1R90)
* [Give a way to update onboarding steps in the future](https://github.com/cozy/cozy-proxy/pull/309/files#diff-bafd0c87fdf0cc50bd63327c02fcecc1R119)

__Important points:__
* ```welcome```, ```agreement``` and ```password``` steps are processing __before the login__, so the only one user data used is the ```public_name```.
    * Endpoint to use: ```POST /register```
* Other steps  will require user authentication to request other user informations from the user document (for pre filled features).
    * Endpoint to use: ```PUT /register```

### DIFF
--- a/server/models/user.coffee
+++ b/server/models/user.coffee
@@ -74,13 +74,23 @@ User.getUsername = (callback) ->
 
 
 User.validate = (data, errors = {}) ->
-    if not helpers.checkEmail data.email
+    if data.email and not helpers.checkEmail data.email
         errors.email = localization.t 'invalid email format'
 
-    if not (data.timezone in timezones)
+    if data.timezone and not (data.timezone in timezones)
         errors.timezone = localization.t 'invalid timezone'
 
     return errors
+
+
+User.checkInfos = (data) ->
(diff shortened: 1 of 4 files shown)


server/models/user.coffee
```diff
@@ -74,15 +76,25 @@ User.getUsername = (callback) ->
```
CPatchane: Good point! This part can be better now for the new onboarding. Thanks @gregorylegarec. It's fixed. 👍

gregorylegarec:
Thx @CPatchane !

PR #323, opened 2016-10-18
[password/step3] Add feature 'show password'

misstick:
## Feature
 - [x] Add feature (a toggle button: show/hide),
 - [x] Add styles


client/app/styles/onboarding.styl
```diff
@@ -87,12 +87,18 @@ form
...
+        button.icon
+            background-size: 1em 1em
+            margin: .55em 0
+
+            span
+                display: none
 
         .hide-password
-            background: red
+            background: aqua url(/app/assets/sprites/hide-eye-icon.svg)
```
misstick: @m4dz : this actually doesnt work; can you tell me which path I can use? I looked at webpackconfig but I didn't understand how to do it. Thx

client/app/styles/onboarding.styl
```diff
@@ -5,6 +5,12 @@
 @import "./base/_colors.styl"
 
 
+// Icons
+
+[data-type=password] svg
+    background-color: #4dcec5
```
m4dz: If we need to add custom bacjground-colors behing SVG icons, I prefer to not hadrcode them in CSS, which may stay generics, and prefer to directly inline background-color in a style attribute in the view template (using a model property to store the color)

client/app/styles/onboarding.styl
```diff
@@ -27,13 +33,24 @@ header
             display: block
 
         svg
-            max-width: 100%
-            height:    (200 / 16)em
+            width:          4.5em
+            height:         4.5em
+            border-radius:  50%
+            padding:        1.5em
+            box-shadow:     0 0 1em 0 #ccc
```
m4dz: Colors are all stored in the `_colors.styl` file. Can we use the `$grey-02` here instead of `#ccc` ?

client/app/styles/onboarding.styl
```diff
@@ -27,13 +33,24 @@ header
...
 
     h1
         margin: 0
 
 
+figure[data-type='preset']
```
m4dz: I don't think a `data-type` attribute is useful, it's too specific IMO. We may try to avoid them, and prefer use and `id` if we really need to target a specific isolated element.

client/app/styles/onboarding.styl
```diff
@@ -58,7 +75,51 @@ footer
...
+
+        button.icon
+            background-size: 1em 1em
+            margin: .55em 0
+
+            span
+                display: none
+
+        .hide-password
+            background: aqua url(/app/assets/sprites/hide-eye-icon.svg)
```
m4dz: It can't work here, as CSS doesn't do any require from assets. If we want to do sprinting, we must include the files in the template using a `svg: use(xlink:href)` syntax (which is perfectly valid for an iconic button).

client/app/styles/onboarding.styl
```diff
@@ -58,7 +75,51 @@ footer
...
+
+        .show-password
+            background: red url(/app/assets/sprites/show-eye-icon.svg)
+
+
+label
+    display: block
+    text-transform: uppercase
+    color: gray
```
m4dz: Can we use colors from the `_colors.styl` palette?

client/app/styles/onboarding.styl
```diff
@@ -67,10 +128,15 @@ footer
     text-transform:   uppercase
     color:            white
 
+.errors
+    color: red
```
m4dz: Can we use colors from the `_colors.styl` palette?

client/app/styles/onboarding.styl
```diff
@@ -58,7 +75,51 @@ footer
...
+    display:        inline-block
+    width:          100%
+    padding:        0.75em 0.75em 0.9em
+    border-radius:  4px
+    box-sizing:     border-box;
+
+
 [role=button]
+button
+    border:           0
```
m4dz: Button styles are already in the #321. @gregorylegarec can you review/merge it?

client/app/styles/onboarding.styl
```diff
@@ -67,10 +128,15 @@ footer
...
     color:            white
 
+.errors
+    color: red
 
 
 // BREADCRUMB
 footer
+    button
+        width: 100%
```
m4dz: Button styles are already in the #321. @gregorylegarec can you review/merge it?

client/app/views/templates/_view_step_base.jade
```diff
@@ -1,5 +1,5 @@
 header
-    figure
+    figure(data-type=stepName)
```
m4dz: I prefer to avoid a `data-type` attribute, or, if we really need it, to use an `id` instead (which, in fact, is what it is here).

misstick:
Note : waif for https://github.com/cozy/cozy-proxy/pull/310/ merge
this PR do not work anymore and I don't find why.

client/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "dependencies": {
     "backbone": "1.3.3",
-    "backbone.marionette": "2.4.6",
+    "backbone.marionette": "3",
```
m4dz: I prefer we won't update to Marionette 3 as it is a major update with many changes and we don't have tested it enough.

misstick: I think its rebase issue... I will fix this

misstick:
@gregorylegarec : I have changed from your review of last Thursday; is it OK for you?

misstick:
@m4dz : I'm not sure about my SVG use; can you tell me if its OK?

m4dz:
Ok for me for the SVG use, thanks!

I've got a little doubt about the `from` wrapping, but nevermind, we'll see that later if it really care :).

client/app/views/steps/password.coffee
```diff
@@ -1,24 +1,63 @@
...
 
     # Get 1rst error only
     # err is an object such as:
     # { type: 'password', text:'step empty fields'}
     serializeData: () ->
-        {
+        return Object.assign {}, @serializeInputData(), {
             error:      @error.message if @error
-            stepName:   @model.get 'name'
+            id:         "#{@model.get 'name'}-figure"
```
gregorylegarec: Two questions for @m4dz here (not blocking for the current PR):

1. What do not use jade's region mechanism and template inheritance here ?
`id` and `figureid` don't need to be in Backbone View ? It is pure template logic, isn't it ?

2. Should we use a classname instead of an id for the figure name ? As we use it only for CSS styling.

client/app/views/steps/password.coffee
```diff
@@ -1,24 +1,63 @@
...
+
+
+    isVisible: false
+
+
+    renderInput: =>
+        data = @serializeInputData()
+
+        # Show/hide password value
+        @$('input[name=password]').attr 'type', data.inputType
```
gregorylegarec: Calling $ is pretty expensive. As a good practice we should cache those results in view's inner variables. See: https://ttmm.io/tech/selector-caching-jquery/
This can be done at the end of `render()` method or the first time `renderInput` is called for example.

client/app/views/steps/password.coffee
```diff
@@ -13,17 +13,23 @@ module.exports = class PasswordView extends StepView
     isVisible: false
 
 
+    afterRender: ->
```
gregorylegarec: `afterRender` does not exist in Backbone (and that's a shame), but Marionette provides an `onRender()` method triggered after each call to `render()`.
http://marionettejs.com/docs/v2.4.7/marionette.itemview.html#render--onrender-event

gregorylegarec:
Thanks @misstick 👍

client/app/views/steps/password.coffee
m4dz: For memory:

1. template inheritance and block regiosn are often the best choice, but I don't understand why you mention it here… For the id, I agree doing it in template is a better choice (its template logic), but not a drama :)

2. I prefer classes for style instead of ids (due to specificity) but here, we faced an exception where we're trying to apply a style (the icon background-color) which is specific to one and only one section. So using id seems legit to me.

PR #324, opened 2016-10-19
Finalize Onboarding welcome step

gregorylegarec:

This PR is a little bit huge but it was hard to split it in several parts afterwhile. So sorry about that.

It essentially does :

* Fix somme little typos or issues server-side detected during integration
* Add a `save()` method to steps to be able to save content each time a step is completed. It also add related docs.
* Finally, it fixes and simplifies routes management :
  * Synchronizes server and client (this last one was too complex)
  * Makes serveral changes into `Application`
  * User is no more redirected to login page when an username exists in DB
  * In client `Application` every `register` route is now handled by a single method that shows onboarding's current step.

### Technical debt :
To save its data, the welcome step uses `fetch()`, but no polyfill is provided. This can be done with webpack. Also, I encountered difficulties to test this method, so no test is provided at this time.

### DIFF
--- a/server/controllers/authentication.coffee
+++ b/server/controllers/authentication.coffee
@@ -10,16 +10,6 @@ localization = require '../lib/localization_manager'
 passwordKeys = require '../lib/password_keys'
 otpManager   = require '../lib/2fa_manager'
 
-# hardcoded onboarding steps order and slug names
-ONBOARDING_STEPS = [
-    'welcome',
-    'agreement',
-    'password',
-    'infos',
-    'accounts',
-    'ending'
-]
-
 
 getEnv = (callback) ->
     User.getUsername (err, username) ->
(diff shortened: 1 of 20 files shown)


server/controllers/authentication.coffee
```diff
@@ -64,7 +71,8 @@ module.exports.onboarding = (req, res, next) ->
                     # registration mode
                     # TODO: this one is temporary, and need to be removed
                     # when we merge CSS again.
-                    res.render 'index', {env: env, onBoarding: true}
+                    env.currentStep = getCurrentOnboardingStep userData
```
CPatchane: It's better to use on line 63, a specific function to compare arrays (this current way doesn't work). I suggest:
```coffee
isArrayEqual = (array1, array2) ->
    if not (array1 and array2)
        return false
    if array1.length isnt array2.length
        return false
    else
        isElementsEqual = array1.every (elem, i) -> elem is array2[i]
        return isElementsEqual
```

CPatchane:
Great job! Thanks @gregorylegarec 🎉

PR #326, opened 2016-10-24
Unify step template width

gregorylegarec:
As seen with @kelukelu, we should unify the width of all step views to `30em`.

Is this two changes seem ok to you @m4dz ?

### DIFF
--- a/client/app/styles/onboarding.styl
+++ b/client/app/styles/onboarding.styl
@@ -115,9 +115,6 @@ form p
   display: flex
   flex-direction: column
 
-button
-    border: 0
-    width: 100%
 
 
 [role=button]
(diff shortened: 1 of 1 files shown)


client/app/styles/onboarding.styl
```diff
@@ -20,7 +20,7 @@
 
 
 main
-    max-width: (600/16)em
+    width: 30em
```
m4dz: `main` here is a flex item. Forcing `width`causes responsive issues. We should prefer `flex: 0 1 30em`, so `main` won't grow above `30em`, but will properly fot in small screens.

gregorylegarec: Ok, thanks !

PR #328, opened 2016-10-25
Add password complexity feature

CPatchane:
* Add password complexity helper + feature.
* Fix also bug breadcrumb display in password component and add password input placeholder.
* Styles Improvements.

### DIFF
--- a/client/app/views/steps/password.coffee
+++ b/client/app/views/steps/password.coffee
@@ -1,4 +1,5 @@
 StepView = require '../step'
+passwordHelper = require '../../lib/password_helper'
 
 
 module.exports = class PasswordView extends StepView
@@ -8,6 +9,7 @@ module.exports = class PasswordView extends StepView
     events:
         'click .next': 'onSubmit'
         'click [action=password-visibility]': 'onToggleVisibility'
+        'keyup input': 'checkPasswordStrength'
 
 
     isVisible: false
@@ -31,6 +33,14 @@ module.exports = class PasswordView extends StepView
         # Update Button Icon
         @$visibilityIcon.attr 'xlink:href', data.visibilityIcon
 
+    initialize: (args...) ->
+        super args...
+        @passwordStrength = 0
+
+    checkPasswordStrength: ->
+        password = @$('input[name=password]').val()
+        @passwordStrength = passwordHelper.getComplexityPercentage password
+        @$('progress').attr 'value', @passwordStrength
 
     # Get 1rst error only
     # err is an object such as:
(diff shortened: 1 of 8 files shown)


client/app/views/steps/password.coffee
```diff
@@ -32,6 +36,29 @@ module.exports = class PasswordView extends StepView
         @$visibilityIcon.attr 'xlink:href', data.visibilityIcon
 
 
+    initialize: (args...) ->
+        super args...
+        # lowest level is 1 to display a red little part
+        @passwordStrength = {percentage: 1, label: 'weak'}
```
gregorylegarec: Shouldn't the strenght be given by a call to passwordHelper, to be sure to have the same logic as the one implemented in the helper ?

client/app/views/steps/password.coffee
```diff
@@ -32,6 +36,29 @@ module.exports = class PasswordView extends StepView
...
+        @updatePasswordStrength = updatePasswordStrength.bind(@)
+
+
+    updatePasswordStrength= _.throttle( ->
+        password = @$('input[name=password]').val()
```
gregorylegarec: We should use to cache jQuery queries, as every call is pretty expensive. (See http://gregfranko.com/blog/jquery-best-practices/). We can to something like
```coffeescript
@passwordInput ?= @$('input[name=password')
```
and then work only with `@passwordInput`.

client/app/views/steps/password.coffee
```diff
@@ -32,6 +36,29 @@ module.exports = class PasswordView extends StepView
...
+        @passwordStrength = passwordHelper.getStrength password
+
+        if @passwordStrength.percentage is 0
+            @passwordStrength.percentage = 1
+        @$('progress').attr 'value', @passwordStrength.percentage
+        @$('progress').attr 'class', 'pw-' + @passwordStrength.label
+        @$inputPassword.removeClass('error')
```
gregorylegarec: Same remarks about jQuery queries caching.

client/app/views/steps/password.coffee
CPatchane: Good point! 👍

CPatchane:
@gregorylegarec: For the two first points, I just pushed the transfix sources and translated them, that should be fixed.

For the last point, I think the password is too short it's why it's too weak. The strenght is also computed according to the password length. But you can also review the password strength helper here in order to be sure.

client/app/views/steps/password.coffee
CPatchane: You're right, it's more maintainable. I'll fix it.

gregorylegarec:
Thanks @CPatchane 👍

PR #329, opened 2016-10-25
Better password errors + fix handleSaveError

CPatchane:
PR based on #328.

* Better errors handling and UX in password component and fix bugs.
* Fix bug about [handleSaveError](https://github.com/cozy/cozy-proxy/pull/329/files#diff-753b070baa8f3302572f4381d40ac8daR100) step method.


client/app/views/steps/password.coffee
```diff
@@ -8,15 +10,22 @@ module.exports = class PasswordView extends StepView
...
+    onRender: (args...) ->
+        super args...
         @$inputPassword = @$('input[name=password]')
         @$visibilityButton = @$('[action=password-visibility]')
         @$visibilityIcon = @$('.icon use')
+        @$errorContainer=@$('.errors')
+        if @error
+            @handleError(@error)
+        else
+            @$errorContainer.hide()
```
misstick: @CPatchane : shouldnt you handle this into template file with a condition `if error`?

client/app/views/steps/password.coffee
```diff
@@ -34,12 +34,14 @@ module.exports = class PasswordView extends StepView
         # Update Button Icon
         @$visibilityIcon.attr 'xlink:href', data.visibilityIcon
 
+
```
misstick: be careful of empty line that add some noise on commits

client/app/views/steps/password.coffee
```diff
@@ -71,7 +103,16 @@ module.exports = class PasswordView extends StepView
...
+        @$errorContainer.html(t(@error))
+        @$errorContainer.show()
+
 
     onSubmit: (event)->
         event?.preventDefault()
-        @model.submit @getDataFromDOM()
+        if @passwordStrength.label is 'weak'
+            @$inputPassword.addClass('error')
+            @handleError("password too weak")
```
misstick: why not handle this into `config/password` file?

client/app/lib/onboarding.coffee
```diff
@@ -98,8 +98,7 @@ class Step
 
     # Error handler for save() call
     handleSaveError: (err) =>
-        error = Object.values err.errors
-        throw new Error error
+        throw new Error err.error
```
misstick: I tested the worklfow when password was too short; server error was formated as `{ error: 'mesage', trans: 'message', errors: 'password is too short' }`. Was this changed in your PR?

CPatchane: According to the [error format use on the server side](https://github.com/cozy/cozy-proxy/blob/development/server/middlewares/errors.coffee#L20) (that I didn't change with my PR #322 ), ```errors``` property could be missing from the server result. It's why I prefer to use here ```err.error``` instead of ```err.errors```. This ```err.error``` message can also be translated after.

client/app/views/steps/password.coffee
CPatchane: If I use the condition in the template instead of here, I would be unable to show an error message directly in the view (because the error container wouldn't be generate in the view).

client/app/views/steps/password.coffee
```diff
@@ -103,16 +103,19 @@ module.exports = class PasswordView extends StepView
...
 
 
     onSubmit: (event)->
         event?.preventDefault()
-        if @passwordStrength.label is 'weak'
-            @$inputPassword.addClass('error')
-            @handleError("password too weak")
+        data = @getDataFromDOM()
+        errors = @model.validate data
+        if errors
```
misstick: I don't understand why you duplicate the model logic into the view? (see https://github.com/cozy/cozy-proxy/blob/development/client/app/models/step.coffee#L23). Maybe is that because Error isnt returned into the view?

The answer is on `application.coffee` https://github.com/CPatchane/cozy-proxy/blob/e6a936e764995cfbb01f17d911627a9395c68552/client/app/application.coffee#L72

````
    initialize: ->
        # ...
        @onboarding.onStepFailed (step, err) => @handleStepFailed(step, err)
        

    #....

    # Update view with error message
    # only if view is still displayed
    # otherwhise dispatch the error in console
    handleStepFailed: (step, err) ->
        if @onboarding.currentStep isnt step
            console.error err.stack
        else
            @showStep step, err

`````

Does this not work?

File: client/app/views/steps/password.coffee [hunk 6]
```diff
@@ -22,6 +22,11 @@ module.exports = class PasswordView extends StepView
         @$visibilityButton = @$('[action=password-visibility]')
         @$visibilityIcon = @$('.icon use')
         @$strengthBar = @$('progress')
+        @$errorContainer=@$('.errors')
+        if @error
+            @handleErrorMessage(@error)
```
gregorylegarec: About the naming here: I am not a big fan of having a call to a method named `handleSomething` when we are not in the situation of an event handling. I suggest:
```Coffeescript
if @error
    @renderError(@error)
// or
if @error
    showError(@error)
```
and a little below:
```Coffeescript
handleErrorMessage: (error) ->
    showError(error)
```

What do you think ?

client/app/views/templates/view_steps_password.jade
```diff
@@ -10,11 +10,10 @@ block region
         label(for="password")=t('step password label')
         button(class='icon', action="password-visibility", title=t(visibilityTxt))
             svg: use(xlink:href=visibilityIcon)
-        input(type=inputType, name="password", id="password" placeholder=t("password placeholder"))
+        input(type=inputType, name="password", id="password" placeholder=t("step password input placeholder"))
     progress.pw-weak(value=passwordStrength.percentage max=100)
 
-    if error
-        p(class="errors")= error
+    p(class="errors")
```
gregorylegarec: I like the `if error` condition in the template, but I understand your need here (being able to update easily a persistent error container. I think we should be able to keep the best of the two worlds by using a Subview. Errors need to be handled too in others steps, so we may update this part soon. Let's go with it for now.

client/app/views/steps/password.coffee
CPatchane: Good point. But actually handleErrorMessage here is the method to render the error message. Maybe we could rename the ```handleErrorMessage``` to ```renderError```?

gregorylegarec:
Thanks @CPatchane 👍

PR #331, opened 2016-10-27
Improve welcome page after product team's review.

gregorylegarec:
* Fix wording
* Skip new onboarding for existing users with password
* Remove temporarily logout link


CPatchane:
Thanks @gregorylegarec !

PR #332, opened 2016-10-27
Fix password strength issue with special chars

gregorylegarec:
* The number of possible special chars was not appropriate
* The special chars regexp was also wrong
* Naming has been changed to clarify algorithm
* Update test

Test have been upgraded to fit more with Product Team's specifications, and also to simplify values testing.


CPatchane:
Thanks @gregorylegarec!

PR #333, opened 2016-10-28
Better server errors handling

CPatchane:


client/app/views/steps/password.coffee
```diff
@@ -103,9 +103,13 @@ module.exports = class PasswordView extends StepView
             onboardedSteps: ['welcome', 'agreement', 'password']
         }
 
-    renderError: (error) ->
+    renderError: (error, useTranslation = true) ->
```
gregorylegarec: What do you think about the idea of letting the server return a translation key instead of a translated message ? It should be less human-friendly but maybe more like what an API could return.

CPatchane: I agree with this approach, but I just wanted to keep the same working as other server errors returned by the proxy. And some error are just not keys, here we can keep server errors to be generic. But I think this is a point that we have to think about in the future.

gregorylegarec: Ok for me.

gregorylegarec:
Thanks @CPatchane. 👍

PR #334, opened 2016-11-02
Add accounts step in onboarding

gregorylegarec:
This PR :
* Integrates HTML/CSS for accounts step
* Fix some minor issue in CSS
* Authenticate user as soon as is password is recorded
* Update server routing mechanism : from now, infos, accounts, and confirmation step are authenticated steps. Two methods have been added on authentification middleware : `isRegistered` and `isNotRegistered` to detect if user has completed onboarding.

### Important
To send credentials to server, our `fetch` call need to have the `credentials` option set to `true`. Otherwise the result of the call is a `401 Unauthorized`. It has been pretty long to understand this part and Passport mechanisms.

### To do
There is still a case to deal with : when user stops registration in accounts step and come back a later. If he is not authenticated anymore, the onboarding will go in a infinite redirection. We'll just have to be able to redirect to login page if user has completed password step, and then redirect to last steps of onboarding once user has logged in.


CPatchane:
Thanks @gregorylegarec. Seems to be great about integration and the confirmation step. I tested your branch and I think that we still have an issue about routing handling authenticated views.
Did you expect to have the login page just after the password saving in order to authenticate the user? If yes, I don't have this behaviour on my own so maybe I did something wrong...

If not, actually the password step currently redirect to the next step (Infos/Accounts) without asking user authentication, but I suggest that should redirect to ```login?next=/register``` after password saving. That will automatically require authentication before continue to the next onboarding steps.

To handle this way, some changes are required here on server side but here are my suggestions that I tested and which work in my environment:

* First, we need an helper to know if the user "authenticatable":

    __server/models/user.coffee__

```coffee
# Returns true if the user data contains enough informations fotr cozy authentication
User.isAuthenticatable = (userData) ->
    if not userData?.onboardedSteps or userData.onboardedSteps.length < 3
        return false
    hasCompletedWelcome = userData.onboardedSteps[0] is 'welcome'
    hasCompletedAgreement = userData.onboardedSteps[1] is 'agreement'
    hasCompletedPassword = userData.onboardedSteps[2] is 'password'

    hasCompletedFirstThreeSteps = hasCompletedWelcome and \
        hasCompletedAgreement and hasCompletedPassword

    # Old user existing before onboarding refactor
    hasPassword = userData and userData.password and userData.salt

    return hasCompletedFirstThreeSteps and hasPassword
```

* Then, if the user is authenticatable, we have to ask the authentication before continue the register steps by using the ```next``` argument of the login feature:

    __server/controllers/authentication.coffee__
    At [Line 44](https://github.com/cozy/cozy-proxy/pull/334/files#diff-5e12fea42396eaa003d0c560f54efa1aR44)

```coffee
# If authenticatable but not registered yet
# -> need authentication to continue next registration steps
if not req.isAuthenticated() and \
    User.isAuthenticatable(userData) and \
    not User.isRegistered(userData)
        res.redirect '/login?next=/register'
# if already registered -> login to cozy home
else if User.isRegistered userData
     res.redirect '/login'
# access onboarding, here the user is not register and:
# - either he is authenticatable and will be authenticated
# - either he is not authenticatable (first three steps)
else
    if userData
        ...
```


* Then, we have to redirect the login (if the user is authenticatable but not registered) with the correct ```next``` argument if it's not already the case:

    __server/controllers/authentication.coffee__
    At [Line 187](https://github.com/cozy/cozy-proxy/pull/334/files#diff-5e12fea42396eaa003d0c560f54efa1aR187)

```coffee
if not User.isRegistered(userData)
    if not User.isAuthenticatable(userData)
        return res.redirect '/register'
    else if req.url isnt '/login?next=/register'
        return res.redirect '/login?next=/register'
```

* The last change is in the password step, on the client side. This step (or the onboarding mecanism) should redirect to ```/login?next=/register``` to require authentication before continue the registration steps.


What do you think about?

server/controllers/authentication.coffee
```diff
@@ -93,14 +93,19 @@ module.exports.saveUnauthenticatedUser = (req, res, next) ->
         User.all (err, users) ->
             return next new Error err if err
             # if existing user document with password -> request rejected
-            if users[0]?.password
+            if users[0]?.password and not requestData.password?
```
CPatchane: I don't really understand why there is another condition here with requestData. In this way, it seems that I would be able to update any other properties of the user document, except password, in an unauthenticated way, isn't it?

server/controllers/authentication.coffee
```diff
@@ -93,14 +93,19 @@ module.exports.saveUnauthenticatedUser = (req, res, next) ->
...
                 error        = new Error 'Not authorized'
                 error.status = 401
                 return next error
             else if users.length
                 users[0].merge userToSave, (err) ->
                     return next new Error err if err
-                    res.status(200).send(result: 'User data correctly updated')
+                    if requestData.password?
+                        return Authentication.authenticate(req, res, next)
```
CPatchane: For my culture, why do you use the authenticate function here? This function is about saving unauthenticated informations only when authorized. If I understand correctly, you want to authenticate the user here when his password is saved? I think we should redirect to the login page with the ```next``` argument in order to require user authentication (see my comment below). What do you think about?

CPatchane:
What do you mean by "authenticate user as soon as his password is recorded"? Did you expect to have the login page just after the password saving in order to authenticate the user? If yes, I don't have this behaviour on my own so maybe I did something wrong...

If not, actually the password step currently redirect to the next step (Infos/Accounts) without asking user authentication, but I suggest that should redirect to ```login?next=/register``` after password saving. That will automatically require authentication before continue to the next onboarding steps.

To handle this way, some changes are required but here are my suggestions that I tested and which work in my environment:
__These changes is maybe out of the scope here but will also complete your todo point.__

* First, we need an helper to know if the user "authenticatable":

    __server/models/user.coffee__

```coffee
# Returns true if the user data contains enough informations fotr cozy authentication
User.isAuthenticatable = (userData) ->
    if not userData?.onboardedSteps or userData.onboardedSteps.length < 3
        return false
    hasCompletedWelcome = userData.onboardedSteps[0] is 'welcome'
    hasCompletedAgreement = userData.onboardedSteps[1] is 'agreement'
    hasCompletedPassword = userData.onboardedSteps[2] is 'password'

    hasCompletedFirstThreeSteps = hasCompletedWelcome and \
        hasCompletedAgreement and hasCompletedPassword

    hasPassword = userData and userData.password and userData.salt

    return hasCompletedFirstThreeSteps and hasPassword
```

* Then, if the user is authenticatable, we have to ask the authentication before continue the register steps by using the ```next``` argument of the login feature:

    __server/controllers/authentication.coffee__
    At [Line 44](https://github.com/cozy/cozy-proxy/pull/334/files#diff-5e12fea42396eaa003d0c560f54efa1aR44)

```coffee
# If authenticatable but not registered yet
# -> need authentication to continue next registration steps
if not req.isAuthenticated() and \
    User.isAuthenticatable(userData) and \
    not User.isRegistered(userData)
        res.redirect '/login?next=/register'
# if already registered -> login to cozy home
else if User.isRegistered userData
     res.redirect '/login'
# access onboarding, here the user is not register and:
# - either he is authenticatable and will be authenticated
# - either he is not authenticatable (first three steps)
else
    if userData
        ...
```


* Then, we have to redirect the login (if the user is authenticatable but not registered) with the correct ```next``` argument if it's not already the case:

    __server/controllers/authentication.coffee__
    At [Line 187](https://github.com/cozy/cozy-proxy/pull/334/files#diff-5e12fea42396eaa003d0c560f54efa1aR187)

```coffee
if not User.isRegistered(userData)
    if not User.isAuthenticatable(userData)
        return res.redirect '/register'
    # avoid looping between register and login redirection
    else if req.url isnt '/login?next=/register'
        return res.redirect '/login?next=/register'
```

* The last change is in the password step, on the client side. This step (or the onboarding mecanism) should redirect to ```/login?next=/register``` to require authentication before continue the registration steps.


What do you think about?

server/controllers/authentication.coffee
gregorylegarec: You've made a good point. As discussed with @poupotte, we will first validate a version of the onboarding where the user is transparently authenticated.
But I agree with you : authenticate at the end of a function supposed to only update the user property is an anti-pattern. Where do you think we may do the authentication ? In a separate method added in the route configuration ? Something like:
```coffeescript
'register/password':
        put: [utils.isAuthenticated, utils.isNotRegistered, auth.saveAuthenticatedUser, auth.authenticate]
```

CPatchane: I really prefer this last approach if that works. It will be better to maintain and I think that matches with the cascading functions pattern of express.

gregorylegarec:
Thanks @CPatchane for your great feedback, I updated the PR with your additionnal code dedicated to login screen when session has expired. Please notice that I updated the code in `User.isAuthenticatable` to deal with only one "last unauthenticated step" to avoid dealing with too much step information in the method. What do you think about this refactoring ?

CPatchane:
@gregorylegarec That's great, I totally agree with this way to keep the code as maintainable as possible 👍

CPatchane:
Thanks @gregorylegarec !

PR #338, opened 2016-11-07
Add confirmation step + redirection at the end of onboarding

CPatchane:
* [x] Confirmation step view
* [x] Redirection using the "Enter" button (`triggerDone` event)
* [x] Translations


client/app/views/templates/view_steps_confirmation.jade
```diff
@@ -5,7 +5,10 @@ block header
     h1=t('step confirmation title')
 
 block region
-    p=t('step confirmation description')
+    p
+        span= t('step confirmation description line 1')
+        br
```
gregorylegarec: I think we should avoid using `<br/>` and prefer `<p>`, even if it does not match exactly the mockups. I asked @kelukelu and he agrees with this point of view.

Have a look to accounts step.

client/app/lib/onboarding.coffee
```diff
@@ -207,7 +207,8 @@ module.exports = class Onboarding
 
     # Trigger a 'done' pseudo-event, corresponding to onboarding end.
     triggerDone: ()->
-        throw new Error 'Not implemented yet'
+        # redirection to the cozy home
+        window.location.replace '/'
```
gregorylegarec: I think that onboarding should not be related to `window` object in any way.

My view :
A listener object like `Application` can subscribe to the "done" pseudo-event like it does for "stepChanged", so we need to add a method `onboarding.onDone(callback)`, which register a callback. Then, `triggerDone` just call the callback.
This way, this is the Application object which knows what to do when the onboarding it's done. It can redirect the brower to a configured url.

client/app/views/steps/confirmation.coffee
```diff
@@ -10,11 +10,29 @@ module.exports = class ConfirmationView extends StepView
...
 
+
     serializeData: ->
         _.extend super,
             id: "#{@model.get 'name'}-figure"
             figureid: require '../../assets/sprites/icon-raised-hands.svg'
+
+
+    renderError: (error) ->
+        console.debug 'renderError', error
```
gregorylegarec: ⚠️  `console.debug`

client/app/lib/onboarding.coffee
CPatchane: Totally agree with you on this point. I changed this part to use this better solution. Thanks @gregorylegarec.

client/app/application.coffee
```diff
@@ -33,9 +33,13 @@ class App extends Application
                 hasValidInfos: ENV.hasValidInfos
             }
 
+            # URL for the redirection when the onboarding is finished
+            @endingRedirection = '/'
```
CPatchane: Oops, you're right. I fixed it.

gregorylegarec:
Nice job @CPatchane.

PR #339, opened 2016-11-07
Fix "ghost user" behaviour

gregorylegarec:
During tests and development sessions, it may happen that a user is authentified in the onboarding process (after filling his password), but does not exists anymore in DB (after a manual removal). In this precise situation, it is not possible anymore to save user's password during the password step.

To avoid that, we decided to throw an error, as this state is not exepcted in a production environment, and could trigger other unexpected error during the onboarding process.

### DIFF
--- a/server/controllers/authentication.coffee
+++ b/server/controllers/authentication.coffee
@@ -74,7 +74,6 @@ module.exports.onboarding = (req, res, next) ->
 # DB. User is still authenticated but does not exist anymore in DB. As it is
 # not an expected state in production, we return a 403 forbidden error.
 module.exports.disallowAuthenticatedUser = (req, res, next) ->
-    console.log req.isAuthenticated()
     if req.isAuthenticated()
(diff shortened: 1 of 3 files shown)


server/controllers/authentication.coffee
```diff
@@ -67,6 +67,21 @@ module.exports.onboarding = (req, res, next) ->
...
 
 
+# Disallow authenticated user
+# Some endpoints, mainly those called by first onboarding steps with post
+# method are not expecting an authentified user. But the situation could happen
+# when developing and after a user authentication followed by user deletion in
+# DB. User is still authenticated but does not exist anymore in DB. As it is
+# not an expected state in production, we return a 403 forbidden error.
+module.exports.disallowAuthenticatedUser = (req, res, next) ->
+    console.log req.isAuthenticated()
```
CPatchane: I think this is a debugging development code line remaining here.

CPatchane:
Thanks @gregorylegarec

PR #345, opened 2016-11-09
Add access to MyAccounts after Accounts onboarding step

gregorylegarec:
Thanks to have a review @CPatchane

In this PR :

### Use of COZY_MYACCOUNTS_URL environment variable ###

If this environment variable is set, cozy-proxy use it as MyAccounts URL. Set to `/app/konnectors` by default. This variable will allow us to efficiently test locally during development and test, without dealing with URL management headache. Just run cozy-proxy with this command :
```terminal
$> PORT=9555 COZY_MYACCOUNTS_URL="http://localhost:9358" npm run watch
```

### Improve base view ###
To easily allow a full-screen iframe without using a dirty `position: absolute` (ping @m4dz), the base view has been a little bit revamped to have its `<main />` element now using the full width and height of the viewport. Its previous CSS properties has been reported into a dedicated `.step-wrapper` element. Side effect : agreement and infos views are upside down but we don't care for now.

### Use a new Marionette view for MyAccounts iframe ###
Instead of a hackish tweak in the Accounts step view for switching to MyAccounts, I prefered to use a brand new Marionnette View. I used Marionnette's clever event bubbling mechanism to make the Application listening to the Accounts view and show the MyAccounts iframe when needed.

### Using elegant flexbox to display the iframe in full-size ###
It speaks by itself. Have a look at the CSS. It's so beautiful. Poke @m4dz.

### DIFF
--- a/client/app/styles/onboarding.styl
+++ b/client/app/styles/onboarding.styl
@@ -234,3 +234,9 @@ footer
         // mark upcoming items in grey
         &[aria-selected=true] ~ li
             background-color  $grey-02
+
+
+// MY ACCOUNTS IFRAME
+.my-accounts-container
+    flex-grow 1
+    border-width 0
--- a/client/app/styles/onboarding.styl
+++ b/client/app/styles/onboarding.styl
@@ -7,12 +7,16 @@
 
 // LAYOUTS
 [role=application]
-    display      flex
     min-height   100vh
     font-family  'Source Sans Pro', sans-serif
 
+// div inserted magically by Marionnette
+main > div
+    display      flex
+    min-height   100vh
 
-main
+
+.step-wrapper
     flex       0 1 30em
     margin     auto
     padding    1em
--- a/client/app/views/steps/accounts.coffee
+++ b/client/app/views/steps/accounts.coffee
@@ -23,7 +23,7 @@ module.exports = class AccountsView extends StepView
 
     onSubmit: (event)->
         event.preventDefault()
-        @model.submit()
+        @triggerMethod 'browse:myaccounts', @model
 
 
     serializeData: ->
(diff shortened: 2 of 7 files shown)


server/controllers/authentication.coffee
```diff
@@ -22,6 +22,8 @@ getEnv = (callback) ->
                 username: username
                 otp:      !!otp
                 apps:     Object.keys require('../lib/router').getRoutes()
+                myAccountsUrl: process.env.COZY_MYACCOUNTS_URL \
+                    or '/app/konnectors/'
```
CPatchane: Is the URL `app/konnectors` working on your side (when konnectors is deployed inside the Cozy so)? When I tried it on my own before and with your branch, that doesn't work. I have to use `/apps/konnectors/` instead.

client/app/styles/onboarding.styl
```diff
@@ -230,3 +234,9 @@ footer
         // mark upcoming items in grey
         &[aria-selected=true] ~ li
             background-color  $grey-02
+
+
+// MY ACCOUNTS IFRAME
+.my-accounts-container
+    flex-grow 1
```
m4dz: 👍

client/app/views/steps/accounts.coffee
```diff
@@ -23,7 +23,7 @@ module.exports = class AccountsView extends StepView
 
     onSubmit: (event)->
         event.preventDefault()
-        @model.submit()
+        @triggerMethod 'browse:myaccounts', @model
```
CPatchane: Great! \o/

gregorylegarec:
Thanks for your feedback on typos @CPatchane, they're fixed now.

CPatchane:
Thanks @gregorylegarec

PR #347, opened 2016-11-14
Add My Accounts access in onboarding's Accounts step.

gregorylegarec:

#### Known problem : ####
I am not sure how to correctly set the href attribute in the accounts step. I let it point to the next property but obviously it is not the case anymore.

To fully accessible, we should need to create a new route for the MyAccounts screen and set it as href in the accounts step link. It implies some work in routing both server side and client side (poke @m4dz).


gregorylegarec:
Sorry, I forgot to rebase my branch.

PR #348, opened 2016-11-14
Add My Accounts access in onboarding's Accounts step.

gregorylegarec:

#### Known problem : ####
I am not sure how to correctly set the href attribute in the accounts step. I let it point to the next property but obviously it is not the case anymore.

To fully accessible, we should need to create a new route for the MyAccounts screen and set it as href in the accounts step link. It implies some work in routing both server side and client side (poke @m4dz).


client/app/views/register/my_accounts.coffee
```diff
@@ -3,6 +3,11 @@
 module.exports = class MyAccountsView extends LayoutView
```
CPatchane: Is it wanted to put this file inside the `register` folder here? I would put it int the `views/subviews/` instead. What do you think?

client/app/styles/onboarding.styl
```diff
@@ -240,3 +242,20 @@ footer
 .my-accounts-container
     flex-grow 1
     border-width 0
+
+.my-accounts-footer
+    font-size 16px
+    height  (48 / 16)em
```
CPatchane: Even we should a global variable here to store this `16px` as font-size, we could use a local variable here for this stylus part. What's your opinion about that?

client/app/views/register/my_accounts.coffee
CPatchane: I understand, but I think `register` is not the correct naming for this folder (a folder to include all views). In prefer `views/onboarding` instead. But I think it's about preferences here.

CPatchane: Great 👍

CPatchane:
Thanks @gregorylegarec!

PR #349, opened 2016-11-16
Minor template and style improvement

gregorylegarec:
Thanks @m4dz to have a review.

This PR prepare the setup of the infos onboarding step.

* Every step was containing a `<form>` tag, even if it was not containing input or button. I used template inheritance and include to offer two base templates for steps.

* There were some incoherences with font size, width, height, etc. I made some fixes to be more compliant with @kelukeku's mockups and to mutualize code between regular steps pages and MyAccounts page.

### DIFF
--- a/client/app/styles/onboarding.styl
+++ b/client/app/styles/onboarding.styl
@@ -87,19 +87,15 @@ a
     text-decoration  none
 
 
-[role=region]
-footer
-    font-size  1.24em
-
-
-
 // FORMS
 
 form
-    font-size  .9em
-
     .line
         position  relative
+
+        // Managing input width with flex
+        display         flex
+        flex-direction  column
 
         button[action="password-visibility"]
             position    absolute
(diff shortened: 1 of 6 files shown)


m4dz:
Sweet 👍 

If I can a suggestion, I'm not especially fan of separating concerns accross multiple templates, only because we need to wrap them in a `<form>` tag, and I ask myself if the form shouldn't be directly put in content, to ensure readability and reassembling concerns, and the footer outside of it. We just have to reference the `form` in the sumbit button using a [form attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-form), and everything will work like a charm 😃

gregorylegarec:
@m4dz Your solution is far better, thanks for the tip !

m4dz:
You're welcome dude

m4dz:
Woohooo 🎉

PR #351, opened 2016-11-17
Better code split between onboarding an login/reset views

CPatchane:
This PR just split the initialisation code to prevent create the full onboarding component when it's not necessary.

### DIFF
--- a/client/app/application.coffee
+++ b/client/app/application.coffee
@@ -48,11 +48,10 @@ class App extends Application
     initializeRouter: () =>
         @router = new Router app: @
         # if onboarding, the pathname will be '/register*'
-        if window.location.pathname.indexOf('/register') is 0
-            @router.route \
-                'register(/:step)',
-                'register',
-                @initializeOnboarding
+        @router.route \
+            'register(/:step)',
+            'register',
+            @handleRegisterRoute
 
 
     # Internal handler called when the onboarding's internal step has just
@@ -77,11 +76,8 @@ class App extends Application
             @showStep step, err.message
 
 
-    # Handler for register route, display onboarding's current step
-    initializeOnboarding: =>
-        # Load onboarding stylesheet
-        AppStyles = require './styles/onboarding.styl'
-        # onboarding steps config
+    # Initialize the onboarding component
+    initializeOnboarding: ->
         steps = require './config/steps/all'
 
         user = {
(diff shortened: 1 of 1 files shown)


client/app/application.coffee
```diff
@@ -58,10 +47,12 @@ class App extends Application
     # Backbone Router
     initializeRouter: () =>
         @router = new Router app: @
-        @router.route \
-            'register(/:step)',
-            'register',
-            @handleRegisterRoute
+        # if onboarding, the pathname will be '/register*'
+        if window.location.pathname.indexOf('/register') is 0
```
gregorylegarec: No need here to test window.location, this mechanism, is already handled by the router.

As we add the `register` route, the `@initializeOnboarding` method will be automatically and transparently called.

client/app/application.coffee
```diff
@@ -87,9 +78,21 @@ class App extends Application
 
 
     # Handler for register route, display onboarding's current step
-    handleRegisterRoute: =>
+    initializeOnboarding: =>
```
gregorylegarec: As this method will be called every time the route will change, i.e. every time a step is validated and the next one "loaded", I think it is not relevant to name it initializeOnboarding, because we do not initalize the onboarding object at every step. We have to do it only one time at the first load.

client/app/application.coffee
```diff
@@ -87,9 +78,21 @@ class App extends Application
...
         AppStyles = require './styles/onboarding.styl'
+        # onboarding steps config
+        steps = require './config/steps/all'
+
+        user = {
+            username: ENV.username,
+            hasValidInfos: ENV.hasValidInfos
+        }
+
+        @onboarding = new Onboarding(user, steps, ENV.currentStep)
```
gregorylegarec: Again, here this code will be called every time the route change, for every step. I think we should to the initialization only once. I propose a code like this one:

```coffeescript````
initializeOnboarding: ->
    steps = require './config/steps/all'

    user = {
        username: ENV.username,
    }

    onboarding = new Onboarding(user, steps, ENV.currentStep)
    onboarding.onStepChanged (step) => @handleStepChanged(step)
    onboarding.onStepFailed (step, err) => @handleStepFailed(step, err)
    onboarding.onDone () => @handleTriggerDone()

    return onboarding


# Handler for register route, display onboarding's current step
handleRegisterRoute: =>
    @onboarding ?= @initializeOnboarding()

    # Load onboarding stylesheet
    AppStyles = require './styles/onboarding.styl'

    currentStep = @onboarding.getCurrentStep()
    @router.navigate currentStep.route
    @showStep(currentStep)

``

CPatchane: You're right, I didn't go far enought in the split of all the onboarding code here, that still called the onboarding initialize functions each time the steps changed. Thanks! 👍

CPatchane:
Thanks for all your feedbacks @gregorylegarec :)

gregorylegarec:
Thanks @CPatchane 👍

PR #353, opened 2016-11-22
Accounts step depends on the presence of the MyAccounts application

goldoraf:
@gregorylegarec It works, but I'm not a big fan of the global ENV variable in the tests. Another approach would be to modify the isActive method so that it takes a second 'installedApps' argument.

### DIFF
--- a/client/app/application.coffee
+++ b/client/app/application.coffee
@@ -82,7 +82,8 @@ class App extends Application
 
         user = {
             username: ENV.username,
-            hasValidInfos: ENV.hasValidInfos
+            hasValidInfos: ENV.hasValidInfos,
+            apps: ENV.apps
         }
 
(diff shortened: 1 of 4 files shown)


gregorylegarec:
@goldoraf Thanks for your work and sorry for review delay.

Yes, It totally agree with you. What do you think about adding the`apps` property in our `user` object ? It could be done here: https://github.com/cozy/cozy-proxy/blob/development/client/app/application.coffee#L83

```coffeescript
user = {
        username: ENV.username,
        apps: ENV.apps
    }
```

So only our Application rely on the `ENV` variable ?

goldoraf:
@gregorylegarec deal! :+1:

gregorylegarec:
Great work @goldoraf, thank you !

PR #355, opened 2016-11-24
Add infos step

gregorylegarec:
Main changes :
* Add default `validate` method in step class which returns a validation object
* Add `fetchData` method into `goToStep` to be able to retrieve data from server.
* `step.submit(data)` now call `step.validate(data)` to secure submission and facilitate validation error management.

Main additions :
* Templates, views, config step for infos step.

Thanks @CPatchane to have a review !


docs/client/onboarding.md
```diff
@@ -244,8 +280,31 @@ let result2 = step.isActive({name: Claudia});
 ##### Parameters
 * `data`: Data to be saved
 
-Submit the step, i.e. try to register it as done. This method should be overriden in config steps to manage specific submits or validation, for steps with forms for example.
-Also, maybe it should return a Promise (return by the save method) to handle correctly remote synchronisation.
+Submit the step, i.e. try to register it as done.
+Returns a Promise, and call `validate(data)` to ensure that data is valid.
+
+#### validate(data)
```
CPatchane: 👍

client/app/views/steps/infos.coffee
```diff
@@ -1,10 +1,139 @@
...
+        'timezone'
+    ]
+
+    initialize: (args...) ->
+        super args...
+
+        # Method used to update submit state when text is typed into inputs.
+        # Throttled to avoid too much calls.
+        @updateSubmitState = _.throttle( =>
+            if @formIsFilled()
```
CPatchane: Also another point, is it not better to test the email format here before enable the submit button?

gregorylegarec:
Thanks for your feedback @CPatchane 👍

client/app/views/steps/infos.coffee
CPatchane: Ok, I agree with your points here. That was just a suggestion but I think we are good. Thanks 👍

CPatchane:
Thanks @gregorylegarec !

PR #356, opened 2016-11-24
Fix current step issue

gregorylegarec:
With infos step and accounts step which may be both inactive, server may pass one of them as current step to client.
This causes an exception, because current step is not in onboarding's active steps list.

This PR fixes this issue, by delegating the decision of the current step to the client. To do so, it passes to it the full onboardedSteps list.


CPatchane:
Thanks!

PR #360, opened 2016-12-05
Fix npm run watch with node 6

nono:


gregorylegarec:
Thanks @nono, does it still work with node 4 ?

nono:
@gregorylegarec yes, it still works with node 4

gregorylegarec:
@nono ok, thanks 👍

PR #363, opened 2016-12-09
Add polyfills to support Array.find and Object.assign

CPatchane:
Add polyfills to support Array.find and Object.assign in older browsers.
Fix tested using Chromium v42

### DIFF
--- a/client/app/initialize.coffee
+++ b/client/app/initialize.coffee
@@ -5,6 +5,8 @@ Sets the browser environment to prepare it to launch the app, and then require
 the application.
 ###
 
+require('./lib/polyfills')
+
 Polyglot = require 'node-polyglot'
 
 app = require './application'
--- a/client/app/lib/polyfills.js
+++ b/client/app/lib/polyfills.js
@@ -1,0 +1,52 @@
+// Array.find polyfill
+if (!Array.prototype.find) {
+  Object.defineProperty(Array.prototype, 'find', {
+    value: function(predicate) {
+     'use strict';
+     if (this == null) {
+       throw new TypeError('Array.prototype.find called on null or undefined');
+     }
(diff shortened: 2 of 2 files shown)


gregorylegarec:
@CPatchane Travis fails with a node4 environment. Can you investigate ?

CPatchane:
@gregorylegarec It's fixed. Good to know: think to empty Travis caches, that can help.

gregorylegarec:
@m4dz We discussed with @CPatchane and we think that we should use the power of webpack to handle polyfills like those ones. What your opinion ? Any recommendation ? (https://www.npmjs.com/package/webpack-polyfills-plugin ? https://www.npmjs.com/package/imports-loader ?

m4dz:
My recommandation for polyfills: always bet on Babel 👅 

In fact, we should use `babel-preset-env` for Babel preset (wich is like an autoprefixer for Babel transpiling) and use its `useBuiltIns` directive that apply polyfills from `babel-polyfill` in the same manner (see https://github.com/babel/babel-preset-env#usebuiltins-boolean). If you want some example, I'll add it to the https://github.com/cozy/cozy-template/pull/17 tomorrow 😄

m4dz:
After discussing, let's say we stuck on manually polyfills as we don't (yet) use Babel on the proxy. Later, when an heterogeneous codebase will come, mixing JS(X) and Coffee, we could use a `babel!coffee` loader to pass output from Coffee-Script to Babel and then use the Babel env capabilities 👍

gregorylegarec:
Ok, so let's wait to add Babel to the project to handle polyfills with it.

I finally succeed in testing the app on browserstack with Chrome 38, it only fail at using `fetch`, so I presume the rest is OK.

PR #367, opened 2016-12-23
Fix /status/activated

clochix:
We are now able to create user when creating the server. /status/registered value was wrong for this users.


clochix:
Ping @aenario

PR #375, opened 2017-05-04
Fix browser autocomplete on passwords

m4dz:


gregorylegarec:
Do we have an idea why Travis is failing ?

gregorylegarec:
Thanks @m4dz 🎶

